Issue #3, opened 2019-12-28
Bug: Navigation bar active link stays on add bookmark


lsxliron:
When navigating between different pages, the navigation bar active link stays on **Add Bookmark**.

The navigation bar should show the current page as active

sissbruecker:
I see that this might be confusing, but the *Add bookmark* button is actually just displayed more prominently because it is the most important action. So when a user looks at the header, this is the action that should stand out, compared to other pages/actions (bookmarklet, settings, logout) that are likely to be rarely used.

So I would prefer not to change this, since having the add bookmark button stand out is more important than highlighting the current page.

lsxliron:
This is especially confusing when viewing on a mobile device.
It is possible to outline the current page menu item and keep the _add bookmark_ untouched

sissbruecker:
I agree that the *Add bookmark* being highlighted in a drop-down menu can be confusing. However I'm still not convinced that adding an outline to the menu item for the current page is the way to go. It seems unnecessary with the other items being rarely used and there would be discrepancy with items like *Logout* which are an action rather than a route.

So instead I moved the *Add bookmark* link out of the mobile menu and changed it into a smaller link button. So now the drop-down only contains non-highlighted items and it's also easier to access the add bookmark page on mobile. Seems like a decent compromise.

PR #5, opened 2019-12-28
Navbar active item is matches the current page

lsxliron:
Fix the navigation bar as described in #3


sissbruecker:
Implemented https://github.com/sissbruecker/linkding/commit/18d4831a5d8269421fb1e6b57a47d3e023589810 as an alternative.

PR #8, opened 2020-02-02
Drop privileges instead of running as root

killgufo:
Hi,
in the current docker container UWSGI runs as root, while there is no need for that.
Instead of dropping privileges and running as user www-data from within docker (which would be preferred IMHO) I run chown on the db file, so that users can transparently upgrade existing setups.


sissbruecker:
Thanks for your contribution @killgufo. I tested this on two installation and had no issues.

Issue #18, opened 2020-05-04
Error importing bookmarks


ericmoon:
Running in a docker from hub.docker.com:
 sissbruecker/linkding:latest | pulled on 2020-04-07 15:16:52

I can't seem to get the import to work.  I've tried exporting from Firefox and Chrome, but always the same error:
An error occurred during bookmark import.

Any pointers to enabling error logs on the docker instance?  If I was running the dev setup, I would enable the debug logs in django settings, but want to avoid that, if possible and keep it within docker.

sissbruecker:
Hey @ericmoon , thanks for the report, I'll try to reproduce when I find some time.

As for logging, I just looked at the code and it looks like the error is caught but not logged 😬. So that would have to be improved as well.

arpitgogia:
Hey @sissbruecker I was able to narrow down the issue to `KeyError` being raised while importing bookmarks, because of the absence of certain attributes in the tag.

I have opened #20 that addresses this

sissbruecker:
The root cause should be fixed thanks to @arpitgogia . I also added some error handling and logging around the importer. I updated the Docker image with the fixes.

I'll close this for now, please reopen if the issue persists.

PR #20, opened 2020-06-06
Fix for Bookmarks Importer

arpitgogia:
Fixes for #18 
- Use a default value if attribute missing from tag
    - Added default '0' for 'toread': This means `bookmark.unread` will default to false
    - Added default '' for 'tags': No tags will be added by default 
    - Added default 'current time(`datetime.now().timestamp()`)' for `date_added`
- Use system's timezone for the parsed timestamp: Earlier the parsed timestamp was not made timezone aware whereas the model's field was timezone aware. This caused a warning while inserting the bookmark into the database. Not sure if this actually caused any problems, I added it because I was able to find it :slightly_smiling_face:


sissbruecker:
Thanks for your looking into this and fixing it @arpitgogia 🙂

PR #34, opened 2020-10-13
Add docker-compose support

LivingWithHippos:
Start your LinkDing instance with `docker-compose up -d`
It still needs the `docker exec -it linkding python manage.py createsuperuser --username=joe --email=joe@example.com` command, it would be nice to automate it [using a .env file](https://docs.docker.com/compose/environment-variables/)

It probably should be mentioned in the README, too.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,11 @@
+version: '3'
+
+services:
+  linkding:
+    container_name: linkding
+    image: sissbruecker/linkding:latest
+    ports:
+      - "9090:9090"
+    volumes:
+      - <your backup data folder>:/etc/linkding/data
+    restart: unless-stopped


krswin (approved):

sissbruecker:
@LivingWithHippos Good idea 👍, but I think it could use some improvements to make it easier to use.

1. The docker-compose config file is currently more of a template (especially the volume configuration), it would make sense to make it configurable so that it can be used out-of-the-box without having to change the docker-compose file itself and setting yourself up for future merge conflicts. I think using an `.env` file is the appropriate way to do this. I would suggest adding an `.env.dist` file which defines all necessary variables (port and volume path so far) and contains default values. Users can then clone the repo, copy the `.env.dist` to `.env` and make changes to it. Then just run `docker-compose up -d`. `.env` should be added to `.gitignore`.
2. There should be a README section which explains how to use this.

LivingWithHippos:
I mistakenly deleted my fork so I think I'll have to close this PR and open another one

So, following the names already in use

- LD_CONTAINER_NAME for the `container_name`
- LD_HOST_PORT for the port
- LD_HOST_DATA_DIR for the data dir

Anyway in the meawhile we can discuss this here

`docker-compose.yml` with default values included (this will work both with and without a .env file)

```yaml
version: '3'

services:
  linkding:
    container_name: "${LD_CONTAINER_NAME:-linkding}"
    image: sissbruecker/linkding:latest
    ports:
      - "${LD_HOST_PORT:-9090}:9090"
    volumes:
      - "${LD_HOST_DATA_DIR:-./linkding_data}:/etc/linkding/data"
    restart: unless-stopped
```

`.env.sample` with default values included

```python
LD_CONTAINER_NAME=linkding
LD_HOST_PORT=9090
LD_HOST_DATA_DIR=/etc/linkding/data
```

Since all the values are "external" I don't think a .env.prod/dev is needed (or a .env at all), maybe if some values that are added to the environment section of the compose in the future it could be more useful

sissbruecker:
@LivingWithHippos That looks very good, would you like to open a new PR? 

Icing on the cake would be to add a section to the `README.md` that quickly explains how to use this with docker-compose (start command) and how a `.env` file can be used to configure the setup 😉

LivingWithHippos:
replaced by #54

PR #54, opened 2020-12-31
Docker compose support

LivingWithHippos:
Added the docker-compose.yml file and some instructions. I wasn't sure about how much text needed to be repeated.

`docker-compose exec linkding python manage.py createsuperuser --username=joe --email=joe@example.com` allows people to run this command even if they changed the container name since `docker-compose exec` checks for the service name in `docker-compose.yml` and not the container name.

Tested with and without the `.env`  file

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,3 +1,3 @@
 LD_CONTAINER_NAME=linkding
 LD_HOST_PORT=9090
-LD_HOST_DATA_DIR=./linkding_data
+LD_HOST_DATA_DIR=./data
--- a/.env.sample
+++ b/.env.sample
@@ -1,0 +1,3 @@
+LD_CONTAINER_NAME=linkding
+LD_HOST_PORT=9090
+LD_HOST_DATA_DIR=./linkding_data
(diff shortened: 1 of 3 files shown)


.env.sample
```diff
@@ -0,0 +1,3 @@
+LD_CONTAINER_NAME=linkding
```
sissbruecker, line 1: Can you please add `.env` to `.gitignore`? Would make it a bit safer IMO, since the file should not be under version control.

sissbruecker:
@LivingWithHippos Looks good, thanks for your contribution!

Looking at the headlines in the README it probably makes sense to restructure them a little bit, but I can take care of that later on.

.env.sample
sissbruecker: Nevermind, it already is. Took my IDE a while to update.

LivingWithHippos: `.env` is already excluded

```
# dotenv environment variables file
.env
```

.env.sample
```diff
@@ -0,0 +1,3 @@
...
+LD_HOST_PORT=9090
+LD_HOST_DATA_DIR=./linkding_data
```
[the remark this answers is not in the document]
LivingWithHippos: The reasoning behind having linkding_data was having multiple docker service  in the same folder but yeah, the user can always change it. Let's use data

sissbruecker:
@LivingWithHippos Perfect, thanks again!

PR #61, opened 2021-01-08
Add !untagged search to find untagged bookmarks

johnbintz:
I want to be able to add bookmarks quickly and organize them later, so I want to be able to list untagged bookmarks. This was the simplest solution I could come up with. Feedback wanted!

* Add .nvmrc file for easier Node development environment setup


bookmarks/templates/bookmarks/tag_cloud.html
```diff
@@ -20,4 +20,6 @@
             {% endfor %}
         </p>
     {% endfor %}
+    <hr />
+    <a href="?q=!untagged">Untagged</a>
```
sissbruecker, line 24: This should modify the existing `q` param rather than overwriting it, like all the other tag links do. That means if the user has searched for `python` and then clicks on `Untagged`, the search should be `python !untagged` instead of `!untagged`. You can check the existing templates on how this is done. 

As stated in the other comment, it might make sense to remove existing tags from the query as well to prevent mixing `!untagged` and `#sometag` in queries.

sissbruecker:
@johnbintz Thanks for the contribution, it's an interesting idea. However I feel that it needs some more work - I'm especially not sure about the search behaviour when combining tags and `!untagged` in the same query. The rest of the comments is just polish.

sissbruecker:
Closing due to inactivity.

PR #89, opened 2021-03-06
Added backup document

mattofr:
Created a small document with some suggestions to backup the database


sissbruecker:
Hey @mattofr , thanks for the contribution. Looks good at first glance. I would probably like a few changes, but I don't think it would be productive to do this using the review process. I'll merge this at some point and clean up the existing readme and reference this document from there.

PR #90, opened 2021-03-07
Multistage Dockerfile

stranger-danger-zamu:
Hey, I really liked the design of the bookmark manager, so I tried to check it out locally. 

I tried to deploy it and it seemed like a lot of intermediate steps going between python and node tooling. I stopped and then tried docker and saw that Docker was also like that. Instead of opening an issue and hoping it'd get tidied up at some point, I instead went ahead and streamlined the build process by modifying the Dockerfile!

Now it'll go through the intermediate steps to build the static files all while building the docker image. So deployment is just:

```bash
docker run -ti --restart always -p 9090:9090 -v $PWD/uwsgi.ini:/etc/linkding/uwsgi.ini --name linkding sissbruecker/linkding
```

As for the changes themselves, the Dockerfile now utilizes mutlistage builds. This both reduces final image size (by 100 MB!) and eats the itermediate build steps that were present before.
*  The first stage `NODEBUILD` grabs package.json, and (eventually) runs `npm run build`.
*  The second stage `PYTHONBUILD` copies the output from `NODEBUILD` and does the python side of things. Additionally for space savings, it utilizes a virtualenv to install all the python dependencies and allow them to be copied out while leaving the build time dependencies (such as libsass-dev, libpcre3-dev,  and mime-support) out of the final image.
*  Finally, the last stage `FINAL` copies the final files from `PYTHONBUILD` along with the `mime.types` required, then runs the bootstrap script. 

When the image runs, it calls uwsgi from the virtualenv and everything is located correctly. Creating an adminuser is still another subsequient `docker exec -ti linkding /venv/bin/python manage.py createsuperuser --username=$USER --email=$EMAIL`  command, but that's not too bad overall.

Hope this helps!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,45 +1,52 @@
-FROM node:current-alpine AS NODEBUILD
-# copy directory to /opt and work inside it
-WORKDIR /opt
-COPY . /opt
-# run the npm portion of build-static script
-RUN rm -rf static && mkdir -p static/
+FROM node:current-alpine AS node-build
+WORKDIR /etc/linkding
+# install build dependencies
+COPY package.json package-lock.json ./
 RUN npm install -g npm && \
-    npm install && \
-    npm run build
+    npm install
+# compile JS components
+COPY . .
+RUN npm run build
 
 
-FROM python:3.9-slim AS PYTHONBUILD
-# install missing system deps
-RUN apt-get update && apt-get -y install build-essential mime-support libpcre3-dev libsass-dev mime-support
-# set working dir
+FROM python:3.9-slim AS python-base
+RUN apt-get update && apt-get -y install build-essential
 WORKDIR /etc/linkding
+
+
+FROM python-base AS python-build
+# install build dependencies
+COPY requirements.txt requirements.txt
+RUN pip install -U pip && pip install -Ur requirements.txt
+# run Django part of the build
+COPY --from=node-build /etc/linkding .
+RUN python manage.py compilescss && \
+    python manage.py collectstatic --ignore=*.scss && \
+    python manage.py compilescss --delete-files
+
+
+FROM python-base AS prod-deps
 COPY requirements.prod.txt ./requirements.txt
-# make a virtualenv, upgrade pip & wheel, install prod requirements
-RUN mkdir /venv && \
-    python -m venv --upgrade-deps --copies /venv && \
-    /venv/bin/pip install --upgrade pip wheel && \
-    /venv/bin/pip install -Ur requirements.txt
-# run the python half of build-static script
-COPY --from=NODEBUILD /opt/ .
-RUN /venv/bin/python manage.py compilescss && \
-    /venv/bin/python manage.py collectstatic --ignore=*.scss && \
-    /venv/bin/python manage.py compilescss --delete-files
+RUN mkdir /opt/venv && \
+    python -m venv --upgrade-deps --copies /opt/venv && \
+    /opt/venv/bin/pip install --upgrade pip wheel && \
+    /opt/venv/bin/pip install -Ur requirements.txt
 
 
-FROM python:3.9-slim as FINAL
+FROM python:3.9-slim as final
+RUN apt-get update && apt-get -y install mime-support
 WORKDIR /etc/linkding
-# copy application code from PYTHONBUILD stage
-COPY --from=PYTHONBUILD /etc/linkding  /etc/linkding
-# copy virtualenv from PYTHONBUILD stage
-COPY --from=PYTHONBUILD /venv /venv
-# copy mime.types from PYTHONBUILD stage
-COPY --from=PYTHONBUILD /etc/mime.types /etc/mime.types
-# run bootstrap.sh
-RUN mkdir -p data && \
-    /venv/bin/python manage.py migrate && \
-    /venv/bin/python manage.py generate_secret_key && \
-    chown -R www-data: /etc/linkding/data
-# expose port and run uwsgi server
+# copy prod dependencies
+COPY --from=prod-deps /opt/venv /opt/venv
+# copy output from build stage
+COPY --from=python-build /etc/linkding/static static/
+# copy application code
+COPY . .
+# Expose uwsgi server at port 9090
 EXPOSE 9090
-CMD ["/venv/bin/uwsgi","/etc/linkding/uwsgi.ini"]
+# Activate virtual env
+ENV VIRTUAL_ENV /opt/venv
+ENV PATH /opt/venv/bin:$PATH
+# Run bootstrap logic
+RUN ["chmod", "+x", "./bootstrap.sh"]
+CMD ["./bootstrap.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,45 @@
-FROM python:3.7-slim-stretch
+FROM node:current-alpine AS NODEBUILD
+# copy directory to /opt and work inside it
+WORKDIR /opt
+COPY . /opt
+# run the npm portion of build-static script
+RUN rm -rf static && mkdir -p static/
+RUN npm install -g npm && \
+    npm install && \
+    npm run build
 
-# Install packages required for uswgi
-RUN apt-get update
-RUN apt-get -y install build-essential
-RUN apt-get -y install mime-support
 
-# Install requirements and uwsgi server for running python web apps
+FROM python:3.9-slim AS PYTHONBUILD
+# install missing system deps
+RUN apt-get update && apt-get -y install build-essential mime-support libpcre3-dev libsass-dev mime-support
+# set working dir
 WORKDIR /etc/linkding
 COPY requirements.prod.txt ./requirements.txt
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,22 +1,45 @@
-FROM python:3.7-slim-stretch
+FROM node:current-alpine AS NODEBUILD
+# copy directory to /opt and work inside it
+WORKDIR /opt
```
sissbruecker, line 3: Not sure what else is in `opt` or ever will be there. How about using a more specific folder? Can reuse `/etc/linkding` as in all the other stages.

Dockerfile
```diff
@@ -1,22 +1,45 @@
...
 
-# Install packages required for uswgi
-RUN apt-get update
-RUN apt-get -y install build-essential
-RUN apt-get -y install mime-support
 
-# Install requirements and uwsgi server for running python web apps
+FROM python:3.9-slim AS PYTHONBUILD
+# install missing system deps
+RUN apt-get update && apt-get -y install build-essential mime-support libpcre3-dev libsass-dev mime-support
```
sissbruecker, line 14: mime-support is listed twice.

Dockerfile
```diff
@@ -1,22 +1,45 @@
...
 
-# Expose uwsgi server at port 9090
+FROM python:3.9-slim as FINAL
+WORKDIR /etc/linkding
+# copy application code from PYTHONBUILD stage
+COPY --from=PYTHONBUILD /etc/linkding  /etc/linkding
+# copy virtualenv from PYTHONBUILD stage
+COPY --from=PYTHONBUILD /venv /venv
+# copy mime.types from PYTHONBUILD stage
+COPY --from=PYTHONBUILD /etc/mime.types /etc/mime.types
```
sissbruecker, line 37: Not sure about this. Would it be a big deal to run this as an apt-get in this stage? It should be a new layer in any case, right? Plus, I think it's not needed in the build stage.

Also there is a diff here in that `build-essentials` is missing. Not sure why that was required before - maybe because of the `pip install`? Do you have any insights here?

Dockerfile
```diff
@@ -1,22 +1,45 @@
...
+# run bootstrap.sh
+RUN mkdir -p data && \
+    /venv/bin/python manage.py migrate && \
+    /venv/bin/python manage.py generate_secret_key && \
+    chown -R www-data: /etc/linkding/data
```
sissbruecker, line 42: All of these need to run when the container is created / after the data folder is mounted into the container:
- `/venv/bin/python manage.py migrate` the purpose is to upgrade the database of the installation if the schema changes, which means it needs access to the respective database
- `/venv/bin/python manage.py generate_secret_key` the secret key must be unique per linkding installation, with this change every install would use the same secret
- same for chown I think - this was to guard against permission issues when mounting a data folder into the container

So all of these would need to run as part of CMD. I guess the idea was to unify the Docker setup a bit, but with the restrictions above I think keeping the separate (and well documented) bootstrap.sh is the better trade-off.

requirements.prod.txt
```diff
@@ -21,3 +21,4 @@ soupsieve==1.9.2
 sqlparse==0.3.0
 urllib3==1.25.3
 uWSGI==2.0.18
+libsass
```
sissbruecker, line 24: Maybe confusion around the wording. Prod. dependencies are supposed to be the modules that the app actually needs at runtime. libsass is only needed during build. I guess there would need to be a second pip install at some point.

sissbruecker:
@stranger-danger-zamu Thanks for the initiative. I like the general idea of making the actual build reproducable, ATM all this stuff happens locally in my environment. But we need to figure out a few things.

Apart from the comments above I wonder about installing the dependencies in one stage and then copying everything into the final stage. The issues I see here:
- build stage installs build dependencies, not all of which are needed in the final stage / at runtime
- this also leads to having to reference everything through `/venv/...` since no paths are set

I think it would be saner to do a second pip install in the final stage with the production dependencies. It might add some overhead in terms of image size, but probably not too much. I would guess the actual large improvement in image size here comes from removing `build-essentials` from the final stage. If we can rule out that this is not needed for anything in the final stage then the improvements would probably still be substantial.

Dockerfile
sissbruecker: RE: build-essentials - apparently required for installing uwsgi.

sissbruecker:
OK, looks like `pip install uwsgi` builds it from source or at least requires a C compiler. Another option that I just gave a quick test with the current Dockerfile:
- remove uwsgi from `requirements.prod.txt` (as well as some other crap I apparently forgot to remove which requires a compiler)
- don't install build-essentials
- install uwsgi through apt-get instead

That alone reduces the image size from 369MB to 213MB. That is probably similar to your results @stranger-danger-zamu ?

Dockerfile
stranger-danger-zamu: `/opt` is empty unless it's filled by the user, feel free to change the approrpaite paths to `/etc/linkding`, `/app`, or even `/workdir`.

Dockerfile
stranger-danger-zamu: Whoops. I was reducing intermediate build steps and didn't catch this one.

Dockerfile
stranger-danger-zamu: This is all personal preference. I always package into virtualenvs to separate concerns (and having better reproducability between dev and prod). And yeah, `build-essential` is for compiling uwsgi when installed via pip.

Dockerfile
stranger-danger-zamu: Ah, I totally forgot about the django migrations and django-generate-secret-key were idempotent, so you could just replace this with the `bootstrap.sh` script. 
As for the migrations, `/venv/bin/python manage.py migrate` should create the database if the sqlite3 file is missing. But that does require the user deploying the container to run the container mounting the `/etc/linkding/data` directory.

requirements.prod.txt
stranger-danger-zamu: Yes, I forgot to go back in and utilize the other requirements.txt file. I tend to use poetry, so they are in the same file, but defined under different headers. The fix for this would be to move the libsass dependency to the normal requirements.txt file, build the dev virtualenv just to compile the sass, and then have another stage for building the prod virtualenv.

stranger-danger-zamu:
> * build stage installs build dependencies, not all of which are needed in the final stage / at runtime

As far as I can see, python's libsass is the only build dependency that is copied over to the final image, but I didn't look too long at the requirements.txt file after I got it to build. 

> * install uwsgi through apt-get instead

Depending on how you define prod. dependencies and reproducibility, you could argue that `uwsgi` _should_ be compiled (not that I am, just playing devils advocate). Installing uwsgi via apt-get seems like a net benefit. It can hurt reproducibility as the apt repo _could_ upgrade the version on you and _might_ break building the image at some point in the future, but that's probably not a concern here.


> * this also leads to having to reference everything through `/venv/...` since no paths are set

It's my preference to have explicit paths to the virtualenv for tracebacks, but it's also simple to remedy. Just adding a `ENV PATH="/venv/bin:$PATH"` after the COPYs. That should allow for just calling the executables in the virtualenv directly.

> That alone reduces the image size from 369MB to 213MB. That is probably similar to your results @stranger-danger-zamu ?

I have around 221MB with uwsgi and libsass being compiled. Removing libsass and probably cleaning up a few things might be the difference.

sissbruecker:
@stranger-danger-zamu OK thanks for the feedback. How do we proceed here? I think the first two stages look very good, I would only make the Python build stage use the `requirements.txt` which already includes lib-sass. The final stage I would like to:
- install uwsgi from apt-get
- run pip install using `requirements.prod.txt` instead of copying the virtual env from build
- restore the usage of the bootstrap.sh

I could also clean up the `requirements.prod.txt` a little, looks like a few `django-sass-processor` dependencies made it into the prod requirements.

Are you up for the changes to the Dockerfile? Shall I contribute to your PR?

stranger-danger-zamu:
So I took a swing at it, but the way that some of the depedencies are pulled in has dev dependencies required at run time (I think django-sass-processor is the culprit). I've run out of time this week to continue working on this, so feel free to take a swing at it. 

Also since there is a npm stage now, you could just do all the css preprocessing with a node preprocessor. That'd probably get rid of the libsass/django-sass-processor/django-compressor dependencies.

Dockerfile
sissbruecker: Changed to /etc/linkding.

sissbruecker:
@stranger-danger-zamu I went over your solution and added some changes:
- added separate stage for production dependencies
- restored the use of the bootstrap.sh CMD
- added separate layers for copying dependency / requirement files to improve caching

I also cleaned the prod requirements files a bit, there was some stuff that was not necessary.

Could not really get the uwsgi through apt-get approach working, the venv solution was much simpler in the end 👍 Thanks again for the initiative. The build is completely reproducable now with Docker and the uncompressed image size is down to less than 200MB. Downside is that a full uncached multi-architecture build now takes about an hour on my system 😬. But I think it's worth it.

Issue #49, opened 2020-12-30
Add a dark mode


dev-kittens:
It would be nice if this included a dark mode.

I will likely PR this at some point (since it's the one thing between now and me using linkding), though I'm not familiar with Django so we'll see~
Otherwise I could skip making it a config setting and just do it in CSS (using `prefers-color-scheme`)... what would be preferred here?

sissbruecker:
> Otherwise I could skip making it a config setting and just do it in CSS (using prefers-color-scheme)... what would be preferred here?

Not sure about this. Recently I used a website that detected that I use dark mode in macOS and switched automatically, but then had awful styles that were hard to read. IMO a setting is more appropriate to prevent such surprises.

In general I'm OK with the idea. I'm up for making this a setting, if you /someone else can provide a quality stylesheet for a dark mode. Note that the solution would need to integrate into the SCSS system that the application uses - I don't have much experience with theming solutions, so I'm not sure how I would approach this. Probably adding a `.dark-mode { ... }` rule to each `*.scss` file that overrides the necessary bits?

sissbruecker:
Added a dark theme and a setting to switch the theme between light/dark/auto. Default is auto which uses prefers-color-scheme to determine the theme.

Export33:
The dark themes looks great. Thank you!

joel72265:
Thank you for the dark theme. 

If someone can share the hex# code of the colors used in the dark theme, I could possibly suggest colors for a possible alternate theme.

PR #92, opened 2021-03-15
Dark Theme

SuiriuS:
This is my proposal for a dark theme such as requested in #49 

I took a very simple approach using the media query [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) to toggle between light and dark. When dark mode is detected, which is determined by the users system settings, the css rules override some of the colors. Browser compatibility is fairly good for any versions from the last 2-3 years. OS compatibility is good as well, Ive tested on Android, Windows, Manjaro, and Fedora with Chrome and Firefox. I havent tested Mac or IOS but I understand they should be on par.
The benefit of using the media query is there's no js changes needed, which made it much easier to do. However if you prefer to have a dedicated switch in the settings then it would be possible to use the same styling as a starting point.

Did not change any of the defaults so the light theme should remain the same.

Color-wise I just used colors off of Github's dark theme and tweaked it until I think it looks reasonable. The colors used are all defined in dark.scss
linkding_dark_1
linkding_dark_2
linkding_dark_3

### DIFF
--- a/bookmarks/styles/base.scss
+++ b/bookmarks/styles/base.scss
@@ -65,14 +65,11 @@ h2 {
     background: $dt-body-bg;
     color: $dt-body-font-color;
   }
-  h2 {
-    color: $dt-gray-color-light;
-  }
   .tab-block {
     border-bottom: solid 1px $dt-border-color;
   }
   .empty {
     background: $dt-bg-color;
-    color: $dt-gray-color-light;
+    color: $gray-color-light;
   }
 }
--- a/bookmarks/styles/base.scss
+++ b/bookmarks/styles/base.scss
@@ -58,3 +58,21 @@ h2 {
 .tab-block {
   border-bottom: solid 1px $border-color;
 }
+
+// Dark Theme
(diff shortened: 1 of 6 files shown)


SuiriuS:
Note that for whatever reason, the uploaded images are actually a lighter shade than they should be. My screenshots look fine so its something that github did to them.

sissbruecker:
@SuiriuS Thanks for the effort, however I decided not to go ahead with your solution and implemented the dark theme myself (#100). I saw several issues with your solution:
- In general the color scheme was not optimal. The contrast for many elements was way too low which made them hard to read. Also many of the colors did not match very well and were way off from the original Spectre.css color scheme.
- I felt your approach of overriding styles per page would not scale very well. Instead I decided to compile two different theme files, which allowed me to just redefine the existing Spectre.CSS variables instead of duplicating existing style rules. That in turn makes most of the components work out-of-the-box, also the ones that are not used yet but might in the future.

Closing.

PR #126, opened 2021-05-17
Fixup scraping

### LINKED ISSUE #148: umlaut/non-ascii characters broken when using bookmarklet (firefox)
When I bookmark a page via the bookmarklet I get broken umlauts/other characters (like é) in the title as well as in the description. As an example:

Bookmarking https://emmikochteinfach.de/pasta-mit-kuerbis-schnell-und-einfach via the bookmarklet produces the title

> "Pasta mit KÃ¼rbis - schnell und einfach gemacht - emmikochteinfach"

and description 

> Das einfache Rezept fÃ¼r Pasta mit KÃ¼rbis. Die schnellen Nudeln mit KÃ¼r...

This does not happen when using the browser plugin (for firefox).

P.S.: That pasta recipe is pretty tasty :-)

kgh02017:
This patchset fixes issue on scraping.

1. Avoid stall on scraping
2. Avoid chracter corruption of scraping some Japanese site

Sincrely,
Taku Izumi

Fixes #148

### DIFF
--- a/bookmarks/services/website_loader.py
+++ b/bookmarks/services/website_loader.py
@@ -34,4 +34,4 @@ def load_website_metadata(url: str):
 
 def load_page(url: str):
     r = requests.get(url, timeout=10)
-    return r.text
+    return r.content
--- a/bookmarks/services/website_loader.py
+++ b/bookmarks/services/website_loader.py
@@ -33,5 +33,5 @@ def load_website_metadata(url: str):
 
 
 def load_page(url: str):
-    r = requests.get(url)
+    r = requests.get(url, timeout=10)
     return r.text
--- a/bookmarks/services/website_loader.py
+++ b/bookmarks/services/website_loader.py
@@ -2,6 +2,7 @@ from dataclasses import dataclass
 
 import requests
 from bs4 import BeautifulSoup
+from charset_normalizer import from_bytes
 
 
 @dataclass
@@ -34,4 +35,10 @@ def load_website_metadata(url: str):
 
 def load_page(url: str):
     r = requests.get(url, timeout=10)
-    return r.content
+
+    # Use charset_normalizer to determine encoding that best matches the response content
+    # Several sites seem to specify the response encoding incorrectly, so we ignore it and use custom logic instead
+    # This is different from Response.text which does respect the encoding specified in the response first,
+    # before trying to determine one
+    results = from_bytes(r.content)
+    return str(results.best())


bookmarks/services/website_loader.py
```diff
@@ -33,5 +33,8 @@ def load_website_metadata(url: str):
 
 
 def load_page(url: str):
-    r = requests.get(url)
-    return r.text
+    try:
+        r = requests.get(url, timeout=3.0)
```
sissbruecker, line 37: Timeout is too low IMO, assuming that the above means 3 seconds. Ideally it would be configurable, until that is possible I would at least suggest 10 seconds. Also just to confirm: is the `timeout` parameter really specified in seconds?

bookmarks/services/website_loader.py
```diff
@@ -33,5 +33,8 @@ def load_website_metadata(url: str):
...
+    except:
+        return ''
```
sissbruecker, line 39: Not sure if it's an improvement to add this fallback, it looks like the site was loaded but was empty. The current error handling is probably not optimal, but on the other hand, why not just let the error bubble up? It will still be handled in `load_website_metadata`.

sissbruecker (changes requested):
@kgh02017 Thanks, makes sense to introduce the timeout here. Added some initial comments below. 

I will need some more time to look into using `content` vs `text`. Meanwhile it would be great if you could explain why using `content` is more appropriate, and provide some example URLs where you noticed the character encoding issues.

kgh02017:
>I will need some more time to look into using content vs text. Meanwhile it would be great if you could explain why using 
>content is more appropriate, and provide some example URLs where you noticed the character encoding issues.

Sample URL is "https://www.mizuhobank.co.jp/retail/products/direct/index.html" (Sorry for Japanese site)
I googleed why character corruption occurs and found that the following link: (sorry for Japanese site too)
 - https://www.lifewithpython.com/2015/07/python-requests-response-body-encoding.html
My understanding is:
 -  it's encoding problem. When scraping site's charaset is not unicode encoded, character corruption happens.
 - request.text manipulates data as "unicode encoded data", while request.content does data as string[].
 So, we can avoid by using requests.content().

bookmarks/services/website_loader.py
kgh02017: I am sorry for my impatience :)
timeout=3.0 means "3 sec timeout"
I'm not familiar with Web technology, so I'm not sure reasonable timeout second.  I think 10 seconds is also good, because
we can avoid "never-end-scraping"

bookmarks/services/website_loader.py
kgh02017: You are right.  try-exception block here is not necessary if timeout exception can be handled outside .

sissbruecker:
From reading the API docs for the request package, it seems using `Response.text` will try to use the encoding from the response headers, and if that is not defined, it will try to determine one using the charset_normalizer package. `Response.content` will simply return the bytes - here I guess that the content will always be read as unicode / the default charset. It seems the problem is that several sites specify an incorrect response encoding, leading `Response.text` to decode the content incorrectly.

I modified the solution to use the charset_normalizer package directly, instead of going through `Response.text`. That means that the encoding that is specified in the response is not respected, and that linkding will always try to determine an encoding that returns useful content, however charset_normalizer determines that. That fixes the issue with some Japansese sites and also solves the issue raised in #148 . We'll have to see if other issues come up with the new solution.

Issue #104, opened 2021-04-04
Show the version in the settings


Export33:
In order to keep track with the latest update (thanks for weekly updates btw) it would be nice to see the version of linkding in the settings.

PR #134, opened 2021-06-22
About section in settings

fumblehool:
Fix: #104 

Screenshot from 2021-06-22 15-36-46


xavier-GitHub76:
Thanks for this commit.

bookmarks/templates/settings/about.html
```diff
@@ -0,0 +1,19 @@
+{% extends "bookmarks/layout.html" %}
+
+{% block content %}
+    <div class="settings-page">
+
+        {% include 'settings/nav.html' %}
+
+        {# Export about #}
+        <section class="content-area">
```
sissbruecker, line 9: Adding a new page / tab for this is a bit much. I think we should add it as a new section to the general settings page (https://github.com/sissbruecker/linkding/blob/master/bookmarks/templates/settings/general.html), below the existing sections.

bookmarks/templates/settings/about.html
```diff
@@ -0,0 +1,19 @@
...
+            <h2>About</h2>
+            <p>Version: {{ version }}</p>
+            <p>
+                Code: <a href="https://github.com/sissbruecker/linkding/"
+                    class="btn btn-link btn-sm" target="_blank">GitHub</a>
```
sissbruecker, line 14: Please remove the class names here. This should just be a regular link.

bookmarks/views/settings.py
```diff
@@ -95,3 +102,11 @@ def _find_message_with_tag(messages, tag):
     for message in messages:
         if message.extra_tags == tag:
             return message
+
+def _get_app_version():
```
sissbruecker, line 106: We can optimize this to run only once. Instead of using a function, I think we could put this into the scope of the module and store the result in a `app_version` variable. So something like:
```python
...

try:
    with open("version.txt", "r") as f:
        app_version = f.read().strip("\n")
except Exception as exc: 
    logging.exception(exc)
    pass

#... Remaining code in the code

@login_required
def general(request):
    if request.method == 'POST':

```

sissbruecker (commented):
Thanks for the contribution! Added some comments that I would like adressed.

Issue #124, opened 2021-05-14
Bookmark import accepts empty URL values


sissbruecker:
The following bookmark tag gets imported successfully, although no URL is specified:
```
    <DT><A HREF="" ADD_DATE="1616337559" PRIVATE="0" TOREAD="0" TAGS="tag1">test title 1</A>
```
This should never happen, as the URL should be required.

This seems to indicate that the `Bookmark` model does not validate the URL field as required, so the issue is probably there. This issue could also affect the REST API.

PR #135, opened 2021-06-22
Added Url validation for on bookmark save

fumblehool:
Fix: #124

### DIFF
--- a/bookmarks/models.py
+++ b/bookmarks/models.py
@@ -1,3 +1,4 @@
+import logging
 from typing import List
 
 from django import forms
@@ -78,6 +79,13 @@ class Bookmark(models.Model):
     def __str__(self):
         return self.resolved_title + ' (' + self.url[:30] + '...)'
 
+    def save(self, *args, **kwargs):
+        try:
+            self.full_clean()
+            return super(Bookmark, self).save(*args, **kwargs)
+        except Exception as exc:
+            logging.exception(exc)
+            raise exc
 
 class BookmarkForm(forms.ModelForm):
     # Use URLField for URL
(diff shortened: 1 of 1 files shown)


sissbruecker:
Thanks for tackling this, but I don't understan how this change would address the issue. The problem in #124  was that it seems to be possible to save a `Bookmark` instance that has no URL or an empty URL. In that case I would assume that the validation for the `url` field is not set up correctly? Or maybe the validation is not triggered as part of the import?

fumblehool:
`self.full_clean()` is called on `.save()` to validate fields.
But as discussed here - https://stackoverflow.com/questions/4441539/why-doesnt-djangos-model-save-call-full-clean ,
in our case, url validation is not happending.

In this commit, I'm overriding `.save()` to validate.
One other way will be to setup `pre-save signal` like this - 
```
@receiver(pre_save)
def pre_save_handler(sender, instance, *args, **kwargs):
    instance.full_clean()
```

sissbruecker:
Thanks for the explanation. To summarize, `full_clean` needs to be called manually when not using Model Forms. I think in that case I would prefer to do just that - call `full_clean` manually before saving a bookmark in the import. Changing the way the save method works seems like a common approach as well, but I think it has downsides:
- it's different from how Django models work out of the box
- according to the SO thread linked above it can have side effects
- it seems overblown for this problem, which only seems to occur in the import ATM (I tested the REST API, which runs validation automatically as well)

As such I would suggest a different solution:
- call `full_clean` before saving in the bookmark in this line: https://github.com/sissbruecker/linkding/blob/master/bookmarks/services/importer.py#L60
- add a test case that verifies that bookmarks with empty URLs are not imported - it seems there are no test cases for the importer itself yet 😬. However there are some integration tests for the import in https://github.com/sissbruecker/linkding/blob/master/bookmarks/tests/test_settings_import_view.py. I think for now it would be OK to add another test for bookmark validation to those.

Feel free to leave this as is. With the solution outlined, I should be able to make the change myself rather quickly.

sissbruecker:
Closing in favor of https://github.com/sissbruecker/linkding/pull/149

PR #138, opened 2021-07-30
Support running on OpenShift as an arbitrary user

crahan:
I recently deployed linkding to an OpenShift environment which required updating the permisisons of the `/etc/linkding` folder to make them group-writable. As per the OpenShift documentation (https://docs.openshift.com/container-platform/4.8/openshift_images/create-images.html), containers on OpenShift do not run as root but as an arbitrary user which is a member of the `root` group:

```
For an image to support running as an arbitrary user, directories and files that are written to by
processes in the image must be owned by the root group and be read/writable by that group.
Files to be executed must also have group execute permissions.
```

Thank you for considering this pull request.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,7 +47,7 @@ EXPOSE 9090
 # Activate virtual env
 ENV VIRTUAL_ENV /opt/venv
 ENV PATH /opt/venv/bin:$PATH
-# Support running container images as an an arbitrary user (e.g., OpenShift)
+# Allow running containers as an an arbitrary user in the root group, to support deployment scenarios like OpenShift, Podman
 RUN ["chmod", "g+w", "."]
 # Run bootstrap logic
 RUN ["chmod", "+x", "./bootstrap.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,6 +47,8 @@ EXPOSE 9090
 # Activate virtual env
 ENV VIRTUAL_ENV /opt/venv
 ENV PATH /opt/venv/bin:$PATH
+# Support running container images as an an arbitrary user (e.g., OpenShift)
+RUN ["chmod", "g+w", "."]
 # Run bootstrap logic
 RUN ["chmod", "+x", "./bootstrap.sh"]
 CMD ["./bootstrap.sh"]


sissbruecker:
Thanks for the initiative. I'm not super-versed with Linux and am not confident about the changes, so I want to double-check some points before moving ahead with this:
1. What is the actual problem that you experience when running on OpenShift? I guess some files were not writable?
2. This older commit (https://github.com/sissbruecker/linkding/commit/ef1c48508273150c177b2641b6c1c61eec306a45) configures the web server running linkding to run as `www-data`. I wonder why that wouldn't be enough in this case?
3. The documentation that you linked actually recommends a different command to setup permissions in the Docker image. Why not use that one?
3. Do you have any other thoughts / considerations about this change? Can you see some negative impact or security issues that may come up as a result of this?

crahan:
Hi, thanks for responding.

1. the issue is with the user account used to run the `bootstrap.sh` script which (among other things) runs `chown -R www-data: /etc/linkding/data`. By default with Docker the user that runs `bootstrap.sh` is `root`. In case of OpenShift or rootless containers, this is a randomly generated user account which is not `root`, but is a member of the `root` group. By making `/etc/linkding` group-writable this allows the randomly generated user which executes `bootstrap.sh` to run the container without issues.
2. `www-data` comes into play later on when the container is run. This is separate from the need to run `bootstrap.sh` which happens earlier on.
3. The command suggested in the documentation sets the group to root and the group permissions for `/etc/linkding` to the same permissions that the owner permissions are set to. So a permission set of `rwxr-xr-x` would be become `rwxrwxr-x` (i.e., match the group permissions to the current user permissions). Running `chmod g+w /etc/linkding` (as suggested in the pull request as part of the Dockerfile) results in the same effect.
4. No additional considerations except for the fact that making `/etc/linkding` group writable is required for the application to work on OpenShift and rootless container setups like Podman. The owner and group for `/etc/linkding` are both already set to `root` without this change. The pull request only makes the directory writable for any user that's a member of the `root` group which does not decrease the container's security.

Dockerfile
```diff
@@ -47,6 +47,8 @@ EXPOSE 9090
 # Activate virtual env
 ENV VIRTUAL_ENV /opt/venv
 ENV PATH /opt/venv/bin:$PATH
+# Support running container images as an an arbitrary user (e.g., OpenShift)
```
sissbruecker, line 50:
```suggestion
# Allow running containers as an an arbitrary user in the root group, to support deployment scenarios like OpenShift, Podman
```

sissbruecker:
Thanks for the detailed explanation, and for the contribution.

Issue #141, opened 2021-08-13
Search by tag is very slow


ulixxe:
Hi,
linkding is a very nice tool for managing bookmarks.
But I find the search by tag so slow that it makes linkding unusable.
I have about 1300 bookmarks and 48 tags. Some tags refer to a large number of bookmarks. For example, the first three most used tags refer to 674, 395, and 255 bookmarks, respectively.
When I search for the first tag (674 bookmarks), linkding hogs with 100% CPU utilization, and after several minutes, I have to restart the docker container to use it again.
I think the search algorithm has some O(n^2) or worst time complexity.
Can it be improved?
Thanks

sissbruecker:
@ulixxe Can you please add your comment to this issue #112 and close this one? Very likely that they are related.

ulixxe:
ok, I will use #112 issue for this problem

PR #142, opened 2021-08-14
changed query on tag search for speedup related to issues #112 and #141

### LINKED ISSUE #112: query with multiple hashtags very slow
I am running linkding in a docker container (following the instructions on this page) on a  Raspberry4 under  a current RaspianOS 32bit with ~500 bookmarks which works very well so far.
The response is quick when searching for a single hashtag, however it can be very long (up to a minute) when combining multiple hashtags in a query (#tag1 #tag2). This is particularly true when that query yields more than just a few results. 
I was wonderung whether the database query can be optimized for speed.
Kind regards, 
(PS great project).

ulixxe:
I tested this change to speed up tag search.
BR

Fixes #112

### DIFF
--- a/bookmarks/queries.py
+++ b/bookmarks/queries.py
@@ -62,46 +62,17 @@ def _base_bookmarks_query(user: User, query_string: str) -> QuerySet:
 
 
 def query_bookmark_tags(user: User, query_string: str) -> QuerySet:
-    return _base_bookmark_tags_query(user, query_string) \
-        .filter(bookmark__is_archived=False) \
-        .distinct()
+    bookmarks_query = query_bookmarks(user, query_string)
+
+    query_set = Tag.objects.filter(bookmark__in=bookmarks_query)
+
+    return query_set.distinct()
 
 
 def query_archived_bookmark_tags(user: User, query_string: str) -> QuerySet:
-    return _base_bookmark_tags_query(user, query_string) \
-        .filter(bookmark__is_archived=True) \
-        .distinct()
+    bookmarks_query = query_archived_bookmarks(user, query_string)
 
-
-def _base_bookmark_tags_query(user: User, query_string: str) -> QuerySet:
-    query_set = Tag.objects
-
-    # Filter for user
-    query_set = query_set.filter(owner=user)
-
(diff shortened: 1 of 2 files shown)


sissbruecker:
Thanks, looks promising. There were test failures, because the new bookmarks queryset used in the intersection was not filtered by the owner. I also changed the code to apply the intersection only if there actually are tags in the query.

I also added some new tests, and fixed some existing tests that check if the query result is empty if the query does not match any bookmarks. As part of that one test started failing: https://github.com/sissbruecker/linkding/pull/142/files#diff-a7e2eb730dc93148ef51f07d9bd7f83f80e941b66e68c01552beee09f65bc4dfR373

Looks like the test also fails in `master`, I'll try to take a look what the issue is here.

sissbruecker:
Yeah, the solution is not correct (but neither was the previous one 🙂).

We have two query sets:
- A: tags that have bookmarks that match all the search terms
- B: tags that have bookmarks that match all the searched tags

And then create an intersection of A and B. The problem is that A might contain results that don't match the searched tags, and B might contain results that don't match the searched terms. If both of these results contain the same tag, it could lead to returning a tag, even through there isn't a bookmark that matches the whole search query, terms + tags combined (there shouldn't be any tags if no bookmark matches the query).

But your solution gave me an idea to re-use the bookmark query as an `IN` filter when querying the tags. The tag query functions then look like this:
```python
def query_bookmark_tags(user: User, query_string: str) -> QuerySet:
    bookmarks_query = query_bookmarks(user, query_string)

    query_set = Tag.objects.filter(bookmark__in=bookmarks_query)

    return query_set.distinct()


def query_archived_bookmark_tags(user: User, query_string: str) -> QuerySet:
    bookmarks_query = query_archived_bookmarks(user, query_string)

    query_set = Tag.objects.filter(bookmark__in=bookmarks_query)

    return query_set.distinct()
```

That solutions seems to be correct, and the performance also seems good. Maybe you can give this a try and check how the performance is in your case?

ulixxe:
Hi,
Yes, it works. The performances are very good!
Thank you!

PR #143, opened 2021-08-14
Have the 'new' form load existing URL

zoredache:
The current behavior of the 'add' form simply overwriting existing bookmarks is pretty annoying. This is particularly true when trying to use using the bookmarklet.

In my opinion it would be ideal to completely merge the new+edit forms. If the 'new' form gets passed an existing URL load all the current details, instead of just allowing a blind overwrite with a warning.

This PR is my attempt at some of the changes you might need to do that.

### DIFF
--- a/bookmarks/templates/bookmarks/new.html
+++ b/bookmarks/templates/bookmarks/new.html
@@ -5,10 +5,18 @@
     <div class="columns">
         <section class="content-area column col-12">
             <div class="content-area-header">
+                {% if 0 == bookmark_id %}
                 <h2>New bookmark</h2>
+                {% else %}
+                <h2>Edit bookmark</h2>
+                {% endif %}
             </div>
+            {% if 0 == bookmark_id %}
             <form action="{% url 'bookmarks:new' %}" method="post" class="col-6 col-md-12" novalidate>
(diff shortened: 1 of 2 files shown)


sissbruecker:
This is an interesting idea, but it needs more work, and some more thought. I might get to this at some point.

feoh:
A wrinkle I just thought about - how would this work if, say someone was submitting a URL through a mobile app/shortcut or an API?

zoredache:
> how would this work if, say someone was submitting a URL through a mobile app/shortcut or an API?

Well it would depend on what we want to happen.

From the server side. I would assume any new URLS being added wouldn't need to change, and deletes URLs wouldn't need to change. Only updates to existing URLs would matter. In the case of duplicate URLS you could just accept the most recent change. Or if you avoid the possibility of accidentally replacing things, you could do something like only accept modification requests if they include the 'date_modified' value of the previous version in the post or something, so you only permitted to modify the record if the client proves it has retrieved the most recent version.  If you use the second option your reject message could return the values for the URL in the database, and the new data, and let the client resolve the conflict and submit an updated request.

sissbruecker:
Closing in favor of https://github.com/sissbruecker/linkding/pull/402

Issue #146, opened 2021-08-20
Error on Google Bookmarks import


ulixxe:
Google will shut down its Bookmarks service (https://www.google.com/bookmarks/) on September 30, 2021. The service won't be available afterward anymore, and users who use it have until then to export their bookmarks to save them.
I tried to import Google Bookmarks, but I got the following error message:
"0 bookmarks were successfully imported.
1393 bookmarks could not be imported. Please check the logs for more details."
Please could you fix it?

PR #147, opened 2021-08-20
Google Bookmarks import fix by converting Windows epoch to Unix epoch.

ulixxe:
I found that Google bookmarks use Windows epoch for date timestamp (starting from 1 January 1601).
This caused an error on utcfromtimestamp function due to date out of range.
I fixed it by checking if the timestamp refers to the Windows epoch (> 11644473600 or equivalently after 1 January 2339 in Unix epoch). In that case, I convert it into Unix epoch.

### DIFF
--- a/bookmarks/services/importer.py
+++ b/bookmarks/services/importer.py
@@ -47,7 +47,13 @@ def _import_bookmark_tag(netscape_bookmark: NetscapeBookmark, user: User):
 
     bookmark.url = netscape_bookmark.href
     if netscape_bookmark.date_added:
-        bookmark.date_added = datetime.utcfromtimestamp(int(netscape_bookmark.date_added)).astimezone()
+        tstamp = int(netscape_bookmark.date_added)
+        if tstamp > 11644473600:
+            # Windows epoch from 1 January 1601 00:00:00
+            bookmark.date_added = datetime.utcfromtimestamp((tstamp-11644473600)/1000000).astimezone()
+        else:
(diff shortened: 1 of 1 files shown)


ulixxe:
This fix resolve issue #146

sissbruecker:
Thanks, it would be nice to support this. Can you please link to some reference for the algorithm / logic that you used? I would like to read some explanation how it works and if there are possible edge-cases. Also the change would require some test cases to verify that both formats are parsed into the correct dates.

ulixxe:
Sorry, I did a mistake on the formula.
I double-checked that Google Bookmarks exports data is in usec, not in seconds. The data format starts from 1 January 1970 (Unix epoch).
So I corrected the formula by converting the timestamp from usec to seconds if an error is raised.
A test case is the following with mixed timestamps. The second one was exported from Google Bookmarks:

<!DOCTYPE NETSCAPE-Bookmark-file-1>
<META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=UTF-8">
<TITLE>Bookmarks</TITLE>
<H1>Bookmarks</H1>
<DL><p>
<DT><A HREF="https://www.bing.com" ADD_DATE="1629892052" PRIVATE="0" TOREAD="1" TAGS="">Bing</A>
<DT><A HREF="https://www.google.com/" ADD_DATE="1629889917805689" PRIVATE="0" TOREAD="1" TAGS="">Google</A>
</DL><p>

PR #152, opened 2021-08-26
archived bookmarks import/export

ulixxe:
This PR aims to export/import archived bookmarks together with all the others.
To distinguish archived bookmarks from not archived ones, I propose to use Netscape-bookmark-file-1 folders, and in particular, a folder named "archive" at the top level folder tree.
To do this, I extended Netscape-bookmark-file-1 parser to manage nested folders too. I defined a new dataclass "NetscapeBookmarkFileData" that stores nested folders and bookmarks. This dataclass is used in importer.py/exporter.py to import/export all bookmarks.

### DIFF
--- a/bookmarks/services/exporter.py
+++ b/bookmarks/services/exporter.py
@@ -1,18 +1,31 @@
 from typing import List
+from datetime import datetime
+
+from django.utils import timezone
 
 from bookmarks.models import Bookmark
+from bookmarks.services.parser import NetscapeBookmarkFileData
 
 BookmarkDocument = List[str]
 
 
-def export_netscape_html(bookmarks: List[Bookmark]):
+def export_netscape_html(netscape_data: NetscapeBookmarkFileData):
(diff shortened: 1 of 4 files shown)


sissbruecker:
Thanks for your effort, but I think the solution is overly complicated for what it needs to accomplish. While I would like to support preserving the archived flag when migrating between linkding instances, I don't see the need to structure archived bookmarks into a folder to add support for folders in the import / export. I would be fine with a solution that just adds an attribute like `archived="true"` to the exported bookmarks, and then parses that attribute on import. Also please note that any new contribution should provide unit tests to verify that the new functionality works, and keeps working, as expected.

As it has been some time I don't know if there is still interest in this, so I'll close the PR for now. Feel free to contribute a new PR with the changes outlined above.

kzshantonu:
I would love to have a way to export archived items separately and independently. I'll just vote here

PR #156, opened 2021-09-01
Allow specifying port through LINKDING_PORT environment variable

cschmatzler:
While Docker containers have their own IP address namespace and port collisions are impossible to achieve, other container solutions share one.
Podman, for example, runs all containers in a pod under one namespace, which results in every port only being allowed to be assigned once. Having a default port in the image that cannot be changed can cause collisions and incompabilities.

This change allows setting a custom port for UWSGI through the environment while keeping the default port as fallback to keep backwards compatibility.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 # Bootstrap script that gets executed in new Docker containers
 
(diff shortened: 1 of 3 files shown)


sissbruecker:
Thanks, renamed the option to `LD_SERVER_PORT` to be more consistent with other options and added docs.

PR #183, opened 2021-12-12
Make bookmarks count column in admin sortable

Fivefold:
The Tag view in django admin has a calculated bookmark count column that is unsortable. This fixes it.


Fivefold:
image

PR #198, opened 2022-02-15
Optional pseudo structure on export

jwezel:
Provide an option to apply a pseudo structure on export. This makes it easier to navigate the bookmarks after import into a browser.

Paths are built from most frequently used tags to least frequently used.

### DIFF
--- a/bookmarks/services/exporter.py
+++ b/bookmarks/services/exporter.py
@@ -1,18 +1,48 @@
-from typing import List
+from collections import Counter
+from functools import reduce
+from itertools import chain
+from typing import Dict, List, Union
 
 from bookmarks.models import Bookmark
 
 BookmarkDocument = List[str]
+Tree = Dict[Union[str, int], Union[Dict, Bookmark]]
 
 
-def export_netscape_html(bookmarks: List[Bookmark]):
-    doc = []
-    append_header(doc)
-    append_list_start(doc)
-    [append_bookmark(doc, bookmark) for bookmark in bookmarks]
-    append_list_end(doc)
+def export_netscape_html(bookmarks: List[Bookmark], apply_pseudo_structure: str):
+    doc: BookmarkDocument = []
+    if apply_pseudo_structure == 'on':
+        append_header(doc)
+        append_list_start(doc)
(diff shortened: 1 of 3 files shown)


sissbruecker:
Kind of questioning the utility of this, specifically how to automatically generate a sensible folder structure from an unstructured set of tags. For example:

Bookmarks:
```
{ tags: "dev, docker" }
{ tags: "dev, css" }
{ tags: "container, docker" }
```

results in:

```
- dev
  - docker
  - css
- docker
  - container 
```

Notice how docker now has two folders. This duplication would likely increase with more bookmarks and more tags.

Another thing is that people **really** like to add lots of tags to their bookmarks, even if they are not reused that often. That would lead to lots of cases where there is a deeply nested structure with just one bookmark in it, and accessing bookmarks requires the user to search through several folders.

I think regardless what approach you take (do not export folders, automatically generate lots of folders), the user will have to clean up the resulting import and restructure the bookmarks into a folder structure that works for them. Automatically generating folders might actually make that more difficult for them. Looking at FF for example, they also have a bookmarks view that is structured by tags. Here you could easily filter for a tag, and then move all of those into a folder of your liking. Seems easier than having to move bookmarks from lots of different folders into one, and then remove the empty folders afterwards.

As an alternative, how about moving this into a separate script that rewrites the HTML export, or into a browser extension? Then you can keep maintaining it, and we can see how many people are actually interested in this. If there is a lot of demand, or feedback on how to improve the algorithm, we can reconsider adding this to the application.

jwezel:
Yes, your example is an edge case (`dev` and `docker` are equally frequent, `dev` was randomly chosen as the "higher up" element in this case).

However, the pseudo structure is not meant as a perfect solution but something to start with when exporting a big number of bookmarks. It doesn't yield the structure the import once had anyway. That's why I made it into an option.

Another solution would of course be to maintain the structure on import and have a hierarchical display in the main page 😉.

> moving this into a separate script that rewrites the HTML export

Not sure what you mean by that. Separate link plus page?

sissbruecker:
As mentioned above, this is not something that I want to adpt for now. Closing.

PR #199, opened 2022-02-16
Import with folders as tags

jwezel:
Provide an option to have folders in bookmark imports serve as tags.

### DIFF
--- a/bookmarks/services/importer.py
+++ b/bookmarks/services/importer.py
@@ -1,12 +1,13 @@
 import logging
 from dataclasses import dataclass
+from typing import List
 
 from django.contrib.auth.models import User
 from django.utils import timezone
 
 from bookmarks.models import Bookmark, parse_tag_string
 from bookmarks.services import tasks
-from bookmarks.services.parser import parse, NetscapeBookmark
+from bookmarks.services.parser import parse, NetscapeBookmark, parse_with_folders
 from bookmarks.services.tags import get_or_create_tags
 from bookmarks.utils import parse_timestamp
 
@@ -20,11 +21,11 @@ class ImportResult:
     failed: int = 0
 
 
-def import_netscape_html(html: str, user: User):
+def import_netscape_html(html: str, user: User, tags_from_folders: str):
     result = ImportResult()
 
     try:
-        netscape_bookmarks = parse(html)
+        netscape_bookmarks: List[NetscapeBookmark] = parse_with_folders(html) if tags_from_folders == 'on' else parse(html)
     except:
         logging.exception('Could not read bookmarks file.')
         raise
--- a/bookmarks/services/parser.py
+++ b/bookmarks/services/parser.py
@@ -1,4 +1,6 @@
 from dataclasses import dataclass
+from html.parser import HTMLParser
+from typing import Dict, List, Optional
 
 import pyparsing as pp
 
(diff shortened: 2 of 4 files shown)


bookmarks/services/parser.py
```diff
@@ -12,18 +14,81 @@ class NetscapeBookmark:
     tag_string: str
 
 
+class BookmarkParser(HTMLParser):
```
sissbruecker, line 17: Not OK with adding a second parser implementation to the codebase. This should either extend the current parser, or replace the existing one. Kind of like the approach here with `HTMLParser`, the implementation might be easier to understand than the current one, and it handles the missing closing tags pretty well.

Also this needs tests. Unfortunately there are no existing tests for the parser itself, but `test_importer.py` could be used to add test cases for converting folders to tags.

bookmarks/services/parser.py
```diff
@@ -12,18 +14,81 @@ class NetscapeBookmark:
...
+            getattr(self, name)()
+        self.tag = None
+
+    def handle_data(self, data):
+        name = f'handle_{self.tag}_data'
+        if name in dir(self):
+            getattr(self, name)(data)
+
+    def handle_start_dl(self, attrs: Dict[str, str]):
+        print('<DL>')
```
sissbruecker, line 44: Remove debug code, or use debug logging

bookmarks/services/parser.py
```diff
@@ -53,7 +118,7 @@ def extract_description(tag):
 bookmark_tag.addParseAction(extract_bookmark)
 
 
-def parse(html: str) -> [NetscapeBookmark]:
+def parse(html: str) -> List[NetscapeBookmark]:
```
sissbruecker, line 121: Out of curiousity, why use `List[...]` rather than `[...]`?

sissbruecker (changes requested):

bookmarks/services/parser.py
jwezel: To my understanding `[...]` is incorrect. The brackets are not type annotations like in C/C++ (`int vec[]` being an array of `int`s) but sort of a parenthesis for specifying generic types, like in a C++ template specifier (`template MyType<class T>`).

bookmarks/services/parser.py
jwezel: OK, going to re-implement the other parser with a subclass of `HTMLParser` and adding a test for it.

bookmarks/services/importer.py
```diff
@@ -20,11 +21,11 @@ class ImportResult:
     failed: int = 0
 
 
-def import_netscape_html(html: str, user: User):
+def import_netscape_html(html: str, user: User, tags_from_folders: str):
```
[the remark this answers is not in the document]
jwezel: The form data for checkboxes is `"on"` (true) / `""` (false). Does Django convert it into `True`/`False` if it's declared `boolean`? I'm not that proficient in Django.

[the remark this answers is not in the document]
sissbruecker: No, my concern is more that this part of the application should be free from frontend semantics. It would be enough to convert into a boolean when calling this function: `request.POST['tags_from_folders'] == 'on'`

bookmarks/services/parser.py
sissbruecker: Cool - just to make sure, the goal is to have a single parser that can handle both cases. I think your existing parser could be parameterized to either create tags from folders, or not.

bookmarks/services/parser.py
```diff
@@ -12,18 +14,81 @@ class NetscapeBookmark:
...
+    def handle_a_data(self, data):
+        print('<A>data:', data)
+        print(' self:', self.href, self.add_date)
+        print(' stack:', self.tag_stack)
+        self.bookmark = NetscapeBookmark(
+            href=self.href,
+            title=data,
+            description=self.description,
+            date_added=self.add_date,
+            tag_string=','.join(self.tag_stack[:-1]),
```
sissbruecker, line 69: I think what's missing here is also applying the tags from the anchor tag itself, currently the explicit tags on the anchor tag are ignored and only the tags created from folders are added.

bookmarks/services/parser.py
```diff
@@ -12,18 +14,81 @@ class NetscapeBookmark:
...
+        self.tag = None
+
+    def handle_data(self, data):
+        name = f'handle_{self.tag}_data'
+        if name in dir(self):
+            getattr(self, name)(data)
+
+    def handle_start_dl(self, attrs: Dict[str, str]):
+        print('<DL>')
+        self.tag_stack.append(None)
```
sissbruecker, line 45: This part, and the related logic in `handle_h3_data` and `handle_a_data` seems a bit convoluted. I think it might be easier to just add a tag to the stack when encountering an `<h3>`, and then to remove a tag from the stack when encountering and `</dl>` - unless the stack is already empty.

sissbruecker:
This has been requested before, and I can see this being useful, so I might still work on this at some point in the future. The parser proposed in this PR has already been adopted in #261.

PR #201, opened 2022-02-21
How-to guide for HTTP Shortcuts app on Android

kzshantonu:


docs/how-to.md
```diff
@@ -20,6 +20,24 @@ Now when you are browsing the web and you want to save the current page as a boo
 
 For more info see here: https://paul.kinlan.me/use-bookmarklets-on-chrome-on-android/
 
+## Using HTTP Shortcuts app on Android
+
+**Note** This allows you to share URL from any app to bookmark it to linkding
+
+Install HTTP Shortcuts from [F-Droid](https://f-droid.org/en/packages/ch.rmy.android.http_shortcuts/) or [Play Store](https://play.google.com/store/apps/details?id=ch.rmy.android.http_shortcuts).
```
sissbruecker, line 27: Not sure if it's a good idea to promote F-Droid at all, but IMO the Play Store link should at least be the first one as that is where people should get their apps unless they really know what they are doing.

docs/how-to.md
```diff
@@ -20,6 +20,24 @@ Now when you are browsing the web and you want to save the current page as a boo
...
+
+Download `linkding_shortcut.zip` from this repository.
```
sissbruecker, line 29: This is likely just a harmless configuration file, however making people download opaque zip files and then recommending them to just put it on their phone seems a bit reckless. 

Would it be too much effort to configure this shortcut manually? Alternatively I would suggest to add a disclaimer to this paragraph that people should verify the contents of the zip before loading it onto their phone to make sure it doesn't contain anything harmful.

Also is there an alternative like an XML or JSON file, that would make it easier to verify the contents.

docs/how-to.md
sissbruecker, line 27: Starting from this paragraph it would make sense to use bullet points, rather than paragraphs. In markdown that looks like this:
```
- Install HTTP shortcuts...
- Download...
- Open HTTP Shortcuts...
...
```

docs/how-to.md
sissbruecker, line 29: Nit: Could link to the actual file here.

sissbruecker (changes requested):
Thanks for the contribution, added a few comments.

sissbruecker:
Looks good 👍

Issue #132, opened 2021-06-10
Option to show URL at all times in default view


onjour:
I'd love to be able to see the full URL of the bookmarks in the default view.

Sometimes the web title isn't informative enough and having the URL as well helps to find the information. 

Thanks

soundneedle:
OMG…esp. in the list of search result. I have no idea what a title is as its “source” without seeing a URL. LinkAce shows that info. Very helpful.

gingerbeardman:
+1 for this

gingerbeardman:
by no means a permanent solution, or the best approach, but I edited a file on my own installation to add this...

https://github.com/sissbruecker/linkding/blob/b93a9fadb637bfb132583836c046887e3f70dd87/bookmarks/templates/bookmarks/bookmark_list.html#L11-L15

becomes:
```
            <div class="title truncate">
                <a href="{{ bookmark.url }}" target="{{ link_target }}" rel="noopener">{{ bookmark.resolved_title }}</a>
            </div>
            <div class="description truncate">
                <span>                                                                                                                          
                <a href="{{ bookmark.url }}" target="{{ link_target }}" rel="noopener">{{ bookmark.url }}</a><br>                               
                </span>                                                                                                                         
                {% if bookmark.tag_names %}
```

and looks like this:
<img width="653" alt="Screenshot 2022-02-22 at 13 26 44" src="https://user-images.githubusercontent.com/49612/155141987-aecf9f70-f402-4199-86cc-0d9800c2a1ff.png">

sebw:
was there ever a merge request for this @gingerbeardman @clach04 @sissbruecker ?

clach04:
> was there ever a merge request for this @gingerbeardman @clach04 @sissbruecker ?

https://github.com/sissbruecker/linkding/pull/219

sissbruecker:
Solved as part of https://github.com/sissbruecker/linkding/pull/365

PR #219, opened 2022-03-18
Issue #132 - show URL in bookmark list

clach04:
Kudos to https://github.com/gingerbeardman

### DIFF
--- a/bookmarks/templates/bookmarks/bookmark_list.html
+++ b/bookmarks/templates/bookmarks/bookmark_list.html
@@ -12,9 +12,6 @@
                 <a href="{{ bookmark.url }}" target="{{ link_target }}" rel="noopener">{{ bookmark.resolved_title }}</a>
             </div>
             <div class="description truncate">
-                <span>
-					<a href="{{ bookmark.url }}" target="{{ link_target }}" rel="noopener" class="text-gray text-sm">{{ bookmark.url }}</a><br>
-				</span>
                 {% if bookmark.tag_names %}
                     <span>
                             {% for tag_name in bookmark.tag_names %}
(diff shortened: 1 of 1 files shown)


sissbruecker:
I'm not totally against displaying the URL in the overview, but at this point it has to be optional. Not everyone wants or needs this, and people like linkding because it's minimal and unobtrusive, so I'm very hesitant to add more elements by default. Apart from that it needs several improvements to the UX and it needs to be covered by automated tests. So in order to adopt this feature I would like the following changes:
- Make displaying the URL optional. It can be enabled in the settings.
- By default the URL is not displayed
- Move the URL into a new line below the title, rather than into the line that contains the description
- Ensure that the line is truncated with elipsis if there is not enough screen space to display the full URL, and ensure that the URL never wraps into a second line or overflows to the right
- Add unit tests to verify that the URL is only visible when the respective setting is enabled, and that it properly displays and links to the URL

sissbruecker (changes requested):
I'm not totally against displaying the URL in the overview, but at this point it has to be optional. Not everyone wants or needs this, and people like linkding because it's minimal and unobtrusive, so I'm very hesitant to add more elements by default. Apart from that it needs several improvements to the UX and it needs to be covered by automated tests. So in order to adopt this feature I would like the following changes:
- Make displaying the URL optional. It can be enabled in the settings.
- By default the URL is not displayed
- Move the URL into a new line below the title, rather than into the line that contains the description
- Ensure that the line is truncated with elipsis if there is not enough screen space to display the full URL, and ensure that the URL never wraps into a second line or overflows to the right
- Add unit tests to verify that the URL is only visible when the respective setting is enabled, and that it properly displays and links to the URL

clach04:
Sounds good, great suggestions! I'll dig into Django a little and see what I can figure out (it's been a while since I used it), could be a while before you see movement on my end.

clach04:
TODO based on https://github.com/sissbruecker/linkding/pull/219#pullrequestreview-922297254

  * Make displaying the URL optional. It can be enabled in the settings.
  * By default the URL is not displayed
  * (/) Move the URL into a new line below the title, rather than into the line that contains the description
  * Ensure that the line is truncated with elipsis if there is not enough screen space to display the full URL, and ensure that the URL never wraps into a second line or overflows to the right
  * Add unit tests to verify that the URL is only visible when the respective setting is enabled, and that it properly displays and links to the URL

sissbruecker:
Closing this as stale for now.

Issue #220, opened 2022-03-18
qrcode support


clach04:
It would be really neat to have qrcode (qr-code) support for URLs.

clach04:
See linked PR for demo (not suitable for merging yet). That implementation uses the web browser to render the qrcode (locally, without dependency on a third party server - which means URLs are not leaked). It does not need additional Python modules - alternative implementation would be implemented via https://github.com/dprog-philippe-docourt/django-qr-code (requires Python 3.7+)

asitemade4u:
I agree on this one.

clach04:
@asitemade4u I pushed a less hacky (but still hacky ;)) version in #221, https://github.com/sissbruecker/linkding/pull/221#issuecomment-1086724502 has screen shots of current progress.

html/js is not my forte though. Its good enough for me to use on my machine but not for merging. If this something you can help with that would be awesome!

PR #221, opened 2022-03-18
Issue 220 qrcode support

clach04:


clach04:
Any interest in this as a feature? This change is not ready for merging! It needs hitting with a pretty stick so that the qrcode is less obtrusive.

This relies on qrcode.min.js from https://github.com/davidshimjs/qrcodejs to be present in the `bookmarks/static/` directory.

clach04:
image
Screen shot

sissbruecker:
@clach04 

> Any interest in this as a feature? 

You would need to make an argument for why this should be added. Who would benefit from something like this, what is the use case? People like linkding because it's minimal and unobtrusive, so the idea is to rather have less features that benefit a lot of people, rather than add all the features that only benefit very few. Especially when that feature would be very obtrusive by taking up a lot of screen space in some manner, or would add extra depencies such as the JS that generates the QR code.

clach04:
@sissbruecker Main use case for me is sharing URLs to friends/coworkers who do not have network access to my instance (or I don't want to give them access) but we have a laptop/TV/phone/tablet in the room to share the qr-code to a phone/tablet.

Shaarli has support for this and the qr code is NOT visible by default, you have to click on a small icon which is a nice way of doing this (my initial hack was to simply just display them all the time super big! :)) Screenshot below in case that's helpful:

image

Shaarli's support is via plugins which leads to one possible idea, any interest in a plugin in architecture for LinkDing? Or would that be too heavyweight? That would change the way one would approach this (and potentially even #219).

Either way, I'm happy running a customized instance of LinkDing so if this isn't useful you've given me the base I need to get the features I want/need :-) Thanks for making this available! I'm currently a reluctant user of Shaarli, it has some of the features I want but it's deployment/storage is not my preference, I like LinkDing's deployment/storage model and it's easier for me to hack on to get the features I want so I'm partway through migrating :)

clach04:
Screen shots of current progress, click to show/hide implemented.

## regular view

image

## clicked

image

sissbruecker:
Thanks for your effort, but I don't want to go ahead with this ATM. Adding another dependency to the app for this is a bit too much IMO, and sharing links is still reasonable easy.

One suggestion I can make is to turn this into a browser extension, and then see if people are interested in this. If this turns out to be popular, then it would be more reasonable to integrate it into the app.

clach04:
On Firefox desktop https://github.com/rugk/offline-qr-code is a good generic (rather than LinkDing specific) browser extension alternative/workaround (doesn't work on mobile though)

sissbruecker:
@clach04 Feel free to mention this in #220 as well, as that is more discoverable for other users.

Issue #187, opened 2021-12-14
Unclear functionality of web archive feature


arminus:
Giving linkdig a spin I really like it so far, but what I don't quite understand is the intended functionality of the web archive feature. So there's a background task which archives every webpage in my database - unless LD_DISABLE_BACKGROUND_TASKS is set to true, right?

- Given the generic name of that flag, what other background tasks are disabled then? 
- If enabled, how often are these tasks / or that task run?
- Apparently, there are downsides to that (see #171 "my IP was banned") would it be reasonable to be able to configure the frequency?
- What is the actual purpose of archiving a bookmarked link if there's apparently no UI integration in linkding? (e.g. optionally open a bookmarked link on http://web.archive.org/) (yeah, I know I can copy&paste the URL into http://web.archive.org)

IMHO, this at least deserves some more explanation in the docs...?

sissbruecker:
The web archive feature creates a snapshot of every bookmarked URL on http://web.archive.org/, also known as Wayback machine. The linkding database does not contain the snapshot itself, but only a link to the snapshot on the web archive.

Regarding your other questions:
- there are no other background tasks ATM. When introducing other tasks, further, more specific, flags can be introduced.
- a task is scheduled when creating a new bookmark, or modifying the URL of an existing bookmark. Same goes for the bookmark import. Additionally when logging in, linkding will schedule archive tasks for every bookmark that does not have a snapshot yet. This covers migration scenarios, where users with an existing database upgrade to a version with that feature.
- as mentioned above, tasks are only scheduled based on user actions. Import and login are special cases that can lead to a large number of tasks. The application will run two tasks in parallel maximum.
- when a bookmark has a snapshot available, the added date turns into a link, and has an additional `∞` character behind it:  ![Bildschirmfoto 2021-12-14 um 21 22 03](https://user-images.githubusercontent.com/357820/146073692-86925436-b3da-43ba-b28e-585755b16dde.png)

Documenting this would definitely be helpful, feel free to contribute to the docs.

arminus:
Thanks, that helps a lot!

gingerbeardman:
thanks for this info!

PR #229, opened 2022-03-28
Issue #187 clarify archive.org feature

clach04:
Use the name in the readme that archive.org use.


clach04:
archive.org was not visible in the readme in html form (e.g. in https://github.com/sissbruecker/) as archive.org was a link. Clarify using the name that IA use.

PR #259, opened 2022-05-18
Add community reference to aiolinkding

bachya:
I started building a linkding workflow for Alfred and realized that instead of burying the linkding API logic in the workflow, it would be helpful (for both that project and others) to create a Python3, async wrapper library. Thus, this PR adds [`aiolinkding`](https://github.com/bachya/aiolinkding) to the community section of the docs.


sissbruecker:
Nice job!

PR #263, opened 2022-05-21
Fake request headers to reduce bot detection

### LINKED ISSUE #262: Getting title not working for articles from nytimes.com
When adding articles from the [New York Times](https://www.nytimes.com/international/), the title of the article is not fetched.
They are using the `<title>`-tag, but with a `data-rh="true"` addition. Maybe this breaks something?

So far, nytimes.com is the only website where I have noticed this behavior.

sissbruecker:
Fixes #262

### DIFF
--- a/bookmarks/services/website_loader.py
+++ b/bookmarks/services/website_loader.py
@@ -34,7 +34,8 @@ def load_website_metadata(url: str):
 
 
 def load_page(url: str):
-    r = requests.get(url, timeout=10)
+    headers = fake_request_headers()
+    r = requests.get(url, timeout=10, headers=headers)
 
     # Use charset_normalizer to determine encoding that best matches the response content
     # Several sites seem to specify the response encoding incorrectly, so we ignore it and use custom logic instead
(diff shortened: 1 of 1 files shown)


bookmarks/services/website_loader.py
```diff
@@ -34,11 +34,22 @@ def load_website_metadata(url: str):
...
 
     # Use charset_normalizer to determine encoding that best matches the response content
     # Several sites seem to specify the response encoding incorrectly, so we ignore it and use custom logic instead
     # This is different from Response.text which does respect the encoding specified in the response first,
     # before trying to determine one
     results = from_bytes(r.content)
     return str(results.best())
+
+
+def fake_request_headers():
```
bachya, line 48: For a future PR, why not make this a constant (since it's contents are always the same)?

sissbruecker: No reason. One article recommended to rotate the user agent, which would have required a function, but I never got that far. It could be changed to a constant.

PR #273, opened 2022-05-27
Add Linkding version endpoint

### LINKED ISSUE #272: Add API call to get linkding version
Now that I've created [`aiolinkding`](https://github.com/bachya/aiolinkding) and [`linkding-cli`](https://github.com/bachya/linkding-cli), I'd like to be able to enforce minimum linkding versions (so that I can notify users if one of those packages requires a newer version than what they have). To that end, it'd be useful to have a simple API call that gets the currently running version.

I attempted a PR for this, but apologies, Django REST Framework is too opaque for my simple mind to understand. 😂

kencx:
Closes #272. Retrieves the current Linkding version at `/api/version` endpoint. 

I hope this doesn't get mistaken for the API's version.

### DIFF
--- a/bookmarks/api/routes.py
+++ b/bookmarks/api/routes.py
@@ -9,6 +9,7 @@ from bookmarks.api.serializers import BookmarkSerializer, TagSerializer
 from bookmarks.models import Bookmark, Tag
 from bookmarks.services.bookmarks import archive_bookmark, unarchive_bookmark
 from bookmarks.services.website_loader import load_website_metadata
+from bookmarks.views.settings import app_version
 
 
 class BookmarkViewSet(viewsets.GenericViewSet,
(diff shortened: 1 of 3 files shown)


bachya:
> I hope this doesn't get mistaken for the API's version.

What about using a different relative endpoint, like `/server`? That way, it could be expanded if other non-version info is desired?

kencx:
That would be good, it would also remove the endpoint from Api Root.

bachya:
I like that!

bookmarks/api/routes.py
```diff
@@ -88,6 +89,16 @@ def get_serializer_context(self):
         return {'user': self.request.user}
 
 
+class VersionViewSet(viewsets.ViewSet,
+                     mixins.ListModelMixin):
+
+    def list(self, request):
```
bachya, line 95: I'm not an expert on Django REST Framework, but will this need to change if this ViewSet could potentially hold other endpoints?

sissbruecker:
I think a more generic `/system` or `/server` endpoint would make sense. Using the `ListModelMixin` seems wrong, as this endpoint will never work with a list of entities. Probably should use some pattern where we can just handle a simple GET request, and then can add nested endpoints when further features are added.

sissbruecker:
Closing in favor of https://github.com/sissbruecker/linkding/pull/392

PR #277, opened 2022-06-05
add keepurl.com online bookmark service to README.md

jiangge:
I love the linkding project, so I build the website keepurl.com for hosting the bookmarks for free.


sissbruecker:
Hey, you probably have the best intentions with this, however I do not feel comfortable with recommending people to sign up (and leave their credentials) with a service hosted by some random person on the internet. We do have some recommended services in the README (PikaPods, fly.io), however in those cases there are companies with some reasonable track record and reputation behind this.
With this service it's not clear what changes were applied to the application, the source code is not available, and just from opening the URL I can see that it loads some trackers from a separate domain that collect data from just opening the page. Doesn't leave the best impression. Going to close this for now.

jiangge:
Hey,

Thanks for your response.

I understand what you worry about, so it's OK.

I've changed the code with adding a tracker  [Umami is a simple, fast, privacy-focused alternative to Google Analytics](https://github.com/umami-software/umami), and translated for Chinese.  

I can add multilingual support for the project in the future, if you feel it is exactly what it needs.

Thank you.

PR #282, opened 2022-06-11
Add apple-touch-icon reference in header

daveonkels:
Recommend adding this reference to support an icon when adding a web app to an iOS homescreen.


sissbruecker:
The favicon didn't work very well as app icon. It seems transparent backgrounds are not allowed, and iOS uses a black background by default. Also the logo was glued right to the edge of the icon. Added a separate icon instead with a white background and a larger padding.

Issue #287, opened 2022-06-26
Duplicated and empty tags on import


wahlm:
I used https://github.com/sbs20/firefox-tags to convert my tree based firefox bookmarks to tagged ones. Then I imported the now tagged bookmarks to linkding and found a lot of duplicated and empty tags in the database (and in the admin panel).

It seems that if a tag is used more than once in the imported file but not yet existing, it is created as many times as it is used. But only one of these duplicated tags gets the connection to the bookmarks. All other tags are not used.

I did some investigation and found out that in importer.py it is checked whether the tag is in the cache, but not, if it is already in the list of tags_to_create.

Although I am no python expert, I did some quick fix and tested it successfully. Below is the patch, but I can provide a PR if wanted/needed.

```
--- a/bookmarks/services/importer.py
+++ b/bookmarks/services/importer.py
@@ -90,9 +90,11 @@ def _create_missing_tags(netscape_bookmarks: List[NetscapeBookmark], user: User)
         for tag_name in tag_names:
             tag = tag_cache.get(tag_name)
             if not tag:
-                tag = Tag(name=tag_name, owner=user)
-                tag.date_added = timezone.now()
-                tags_to_create.append(tag)
+                tag = next((x for x in tags_to_create if x.name == tag_name), None)
+                if not tag:
+                    tag = Tag(name=tag_name, owner=user)
+                    tag.date_added = timezone.now()
+                    tags_to_create.append(tag)
 
     Tag.objects.bulk_create(tags_to_create)
 

```

sissbruecker:
Thanks for taking a look! I think it would make sense to add new tags to the cache (see `TagCache.put`), then the existing condition would already prevent the duplicates. The tag could be added to the cache at the end of this block: https://github.com/sissbruecker/linkding/blob/88fcb422923d5bae80a89ae061d13f46bd200040/bookmarks/services/importer.py#L92-L95

Feel free to open a PR, this should also have a unit test IMO, but I can add that one if you're not comfortable with it.

sissbruecker:
Fixed as part of #289

PR #289, opened 2022-06-28
no duplication of imported tags (#287)

wahlm:
Solves #287 
Yes, the solution you proposed is just simpler and better. In addition I wrote a test to verify the solution.


sissbruecker:
Thanks, replaced the test with a more appropriate one.

PR #300, opened 2022-07-17
Add links to remove tags from current query

jhauris:
Simple addition of links to current tags so they can be removed one-by-one. Helps ease navigation through collection of links.

### DIFF
--- a/bookmarks/styles/bookmarks.scss
+++ b/bookmarks/styles/bookmarks.scss
@@ -106,6 +106,13 @@ ul.bookmark-list {
   }
 }
 
+.current-tag-cloud {
+  a, a:visited:hover {
+    color: $alternative-color;
+    font-style: italic;
+  }
+}
+
 .bookmarks-form {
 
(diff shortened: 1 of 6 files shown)


sissbruecker:
This is a cool idea, but it could use some improvements:
- visuals could be improved by moving the current tags below the "Tags" header, and adding more visual clues that clicking this link will remove the tag. Quick mockup: 
    Bildschirmfoto 2022-07-25 um 09 46 17

- the current tags could be removed from the tag cloud
- to get the current tags from the search query, the `_parse_query_string` function could be exposed and reused
- there's a bug that also lists other search terms as current tags, would likely be fixed by doing the above
- duplicates in the current tags should be removed
- it needs tests

I have some spare time ATM, so I might give this a shot myself.

sissbruecker:
Completed this in #307 using your commit as a base

jhauris:
This is great, thank you for cleaning it up and finishing it.

PR #310, opened 2022-08-02
Order tags in test_should_create_new_bookmark

RoGryza:
This test seems to rely on the tags being queried in a specific order from the database, which is incidentally the case for SQLite.

This came up when I was merging my postgres fork with the latest release [here](https://github.com/sissbruecker/linkding/commit/99582f3d45d0c2d54df66113c019c95c75841c1d#diff-fa1abe4b8e2fd558c4c4450b3bd9b5bf0a255cfc516b8d125fcbf7534d7d1efdR41-R43)


sissbruecker:
Thanks, could you please also change the edit view test:
https://github.com/sissbruecker/linkding/blob/e6718be53b97766d2203eba3b402b5959e86917f/bookmarks/tests/test_bookmark_edit_view.py#L41-L42

RoGryza:
Done

Issue #215, opened 2022-02-25
Share Bookmarks between users


dojo90:
I am currently looking for a solution to manage links from the internet with my girlfriend.

I am thrilled with the performance and the existing integration of Firefox/Chrome! A big praise from my side! 🎉👍

What I am currently missing is the ability to share links between multiple users. Is this feature already included or planned for the future?
Currently we would both have to use the same account. 🤔

sissbruecker:
That's not possible at the moment, but I agree that sharing bookmarks with individual users can be useful, moreso than having publicly visible bookmarks (#70 ). For the most part the application is currently designed to be used by individuals, and my assumption is that the overwhelming majority of use-cases are just that - a single installation is used by a single person.

No plans to work on this for now, but people can always vote (👍) on tickets to give an indication what to focus on when adding new features.

petersem:
@sissbruecker 
Please add url sharing.

This would be very useful, especially if it could be done for shared groups or links.

PR #311, opened 2022-08-04
Add bookmark sharing

sissbruecker:
Allows to share bookmarks with other users, and introduces a new view for shared bookmarks. Bookmarks can be marked as shared in the bookmark form, or through the API. Shared bookmarks are visible to all other authenticated users, but can not be viewed anonymously. Sharing bookmarks with a specific set of users is not possible at the moment. The shared bookmarks view shows shared bookmarks from all users, and allows filtering by search terms, tags from all users, and by a specific user. Shared bookmarks from other users are not shown in the regular bookmark views for active or archived bookmarks.

To avoid cluttering the UI for users that use linkding with a single user, the feature is opt-in and needs to be enabled in the settings first.

Closes https://github.com/sissbruecker/linkding/issues/215


bachya:
FYI, I think the [API docs](https://github.com/sissbruecker/linkding/blob/master/docs/API.md) are missing info on this new `shared` parameter. Happy to put a PR together.

PR #334, opened 2022-09-04
Improve bookmark query performance

sissbruecker:


sissbruecker:
@RoGryza This might be of interest to you. I looked a bit into the query performance of the bookmarks list, and it turns out the optimization of aggregating / concatenating tag names is actually rather slow due to the left outer joins it requires. This change would remove the custom aggregation, which I understand was something you had to work around for the Postgres fork, and instead loads the tags for the displayed bookmarks in a separate query.

PR #368, opened 2022-11-22
Gracefully stop docker container when it receives SIGTERM

### LINKED ISSUE #284: "docker-compose down" always times out
When I run `docker-compose down`, the SIGTERM sent to the container appears to be ignored. Instead, I have to wait 10 seconds (the default timeout) until a SIGKILL is sent to forcefully shut it down.

It's frustrating when I stop (or restart) a number of services and see that linkding is the only one that doesn't shut down right away. Can this be fixed?

mckennajones:
Fix for #284. Useful for when running in docker or kubernetes. 

Two things it fixes:
* Signals were not making their way to the `uwsig` process as it was invoked from the `bootstrap.sh` entrypoint script. By `exec`ing that command, it will receive signals. 
* However, in `uwsig` versions before 2.10, it treats `SIGTERM` a bit oddly. [See here](https://uwsgi-docs.readthedocs.io/en/latest/ThingsToKnow.html):
> Till uWSGI 2.1, by default, sending the SIGTERM signal to uWSGI means “brutally reload the stack” while the convention is to shut an application down on SIGTERM. To shutdown uWSGI use SIGINT or SIGQUIT instead. If you absolutely can not live with uWSGI being so disrespectful towards SIGTERM, by all means enable the die-on-term option. Fortunately, this bad choice has been fixed in uWSGI 2.1

By setting `die-on-term` in the .ini we can avoid this behavior. 

## Testing
Without changes (using latest linkding image):
```
➜  ~/temp/linkding docker-compose up -d
[+] Running 2/2
 ⠿ Network linkding_default  Created                                                 0.0s
 ⠿ Container linkding        Started                                                 0.3s
➜  ~/temp/linkding  docker-compose down
[+] Running 2/2
 ⠿ Container linkding        Removed                                                 0.2s
 ⠿ Network linkding_default  Removed                                                 0.1s
```

With changes:
```
➜  ~/temp/linkding docker-compose up -d
[+] Running 2/2
 ⠿ Network linkding_default  Created                                                  0.0s
 ⠿ Container linkding        Started                                                  0.3s
➜  ~/temp/linkding docker-compose down
[+] Running 2/2
 ⠿ Container linkding        Removed                                                  1.1s
 ⠿ Network linkding_default  Removed                                                  0.0s
```

Fixes #284

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -22,4 +22,4 @@ if [ "$LD_DISABLE_BACKGROUND_TASKS" != "True" ]; then
 fi
 
 # Start uwsgi server
-uwsgi --http :$LD_SERVER_PORT uwsgi.ini
+exec uwsgi --http :$LD_SERVER_PORT uwsgi.ini
--- a/uwsgi.ini
+++ b/uwsgi.ini
@@ -11,6 +11,7 @@ stats = 127.0.0.1:9191
 uid = www-data
 gid = www-data
 buffer-size = 8192
+die-on-term = true
 
 if-env = LD_CONTEXT_PATH
 static-map = /%(_)static=static


sissbruecker:
Thanks, tested the changes and it seems to solve the issue. Also thanks for the detailed write-up.

Was a bit confused by the mention of uwsgi 2.1, which doesn't have a release. I faintly remember that 2.1 was a larger refactoring which was never finished (as of yet). So since upgrading is not an option, this is probably the best solution for now.

PR #371, opened 2022-11-27
Add apache reverse proxy documentation.

jhauris:
Also includes a docker-compose example, and minimal vhost configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,10 +104,23 @@ When using a reverse proxy, such as Nginx or Apache, you may need to configure y
 <details>
 <summary>Apache</summary>
 
-Not tested yet.
-If you figure out a working setup, feel free to contribute it here.
+Apache2 does not change the headers by default, and should not
+need additional configuration.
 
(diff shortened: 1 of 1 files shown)


sissbruecker:
Thanks for testing this! I'm not so sure about maintaining the whole Apache setup from this PR though (Dockerfile, full Apache config + Vhost config). I think it would be enough to just inline the virtual host config into the Readme details section as an example. Alternatively you could put the setup in your own repo, and we can link to that from the Readme. What do you think?

jhauris:
> Thanks for testing this! I'm not so sure about maintaining the whole Apache setup from this PR though (Dockerfile, full Apache config + Vhost config). I think it would be enough to just inline the virtual host config into the Readme details section as an example. Alternatively you could put the setup in your own repo, and we can link to that from the Readme. What do you think?

That makes sense. Let me re-work this PR as you suggest:
* add the virtual host config to the ReadMe
* Create a tag in my repo in the branch with the docker compose setup
* add a link in the ReadMe to the configuration in my repo

Thank you for the suggestion!

jhauris:
This PR now only includes updates to the Readme. If you don't like the link to my repo, we can just cut out the last commit.

sissbruecker:
Looks good, thanks.

Issue #157, opened 2021-09-05
archive.org integration should be configurable (opt-in)


fs111:
With the new release linkding sends all urls to archive.org. While that may be useful to some it is a privacy problem. I am self-hosting my bookmarks to limit the amount of services that know about my bookmarks. With the new feature that is no longer the case.

I assume setting `LD_DISABLE_BACKGROUND_TASKS` to `True` will help, but I that seems broader than just archive.org integration.

Side note: Using a positive term ("Enable") and setting it to `False` is a lot more obvious to most humans, than having a "Disable" flag.

sissbruecker:
Thanks for bringing this up, and I understand your concern. However at this time I would like to keep this feature as opt-out, as I think that the majority of users would expect this feature to work OOTB. I'm open to reconsider this if there is enough data to prove that the majority of users would like to have this disabled by default.

I also think that there is a difference in terms of privacy concerns with an online service where you have an account with bookmarks that is easily accessible by said service, vs. a self-hosted service that would allow to correlate your IP address to your bookmarks, assuming that the IA has an interest in doing this. Again, not saying that your concerns aren't valid, just stating that there are different levels of concerns when talking about self-hosting.

In the meantime you can safely use the `LD_DISABLE_BACKGROUND_TASKS` option to turn off the feature. The option doesn't do anything else at the moment. If another feature comes up that would require background tasks, then we can consider adding a new option just for turning off the snapshot creation.

sissbruecker:
Another alternative I can think of is adding a user setting for this. Then the issue is discoverable and explainable in the UI, and you could easily turn it off without messing with Docker options.

M4NH3X:
Wanted to suggest the same feature but luckily found the issue here. Yes please, a settings in the web-ui itself would be really handy.

I just didn't like the fact that it's enabled by default without giving an opt out option. It was my mistake to not read main readme page carefully. I will re-deploy with `LD_DISABLE_BACKGROUND_TASKS` as the workaround.

Lastly, love the ui and sleekness of this tool. Great work on that and appreciate the project. :+1:

917huB:
I’d like to see the ability to configure a link to a self hosted archivebox install to archive articles locally rather than on cloud.

clach04:
A web-config option would obviously be great, but I personally consider opt-in as the the privacy/safety option.

[...]

PR #381, opened 2022-12-23
refactor: fetch site metadata asynchronously when saving

tecoholic:
## Description

The PR makes fetching the site metadata (title, description) asynchronously when creating and updating the Bookmarks. 

## Problem

I noticed that saving bookmarks using the browser extension on my self-hosted instance was taking way too long, sometimes even 10 seconds.

Sample: **5.92 seconds**
sync_extension_hosted

The core problem seems to be that fetching the URL, parsing the content in BeautifulSoup, setting the metadata, all seem to be happening in the Request-Response cycle of saving a Bookmark.

## Solution

In this PR, the issue is solved by making the `_update_website_metadata` run asynchronously.

## Results

The Bookmark create and update requests happen in a fraction of the time as before. Below are screenshots of some comparisons

All the following tests were performed with linkding running on local machine and the save Wikipedia link was saved.

#### Before - 510ms
|sync|
|-|

#### After - 180ms
|async|
|-|

#### Before - 951ms

|sync_extension|
|-|

#### After - 113ms

|async_extenstion|
|-|

### DIFF
--- a/bookmarks/services/bookmarks.py
+++ b/bookmarks/services/bookmarks.py
@@ -1,5 +1,6 @@
 from typing import Union
 
+from background_task import background
 from django.contrib.auth.models import User
 from django.utils import timezone
 
@@ -17,14 +18,14 @@ def create_bookmark(bookmark: Bookmark, tag_string: str, current_user: User):
         _merge_bookmark_data(bookmark, existing_bookmark)
         return update_bookmark(existing_bookmark, tag_string, current_user)
 
-    # Update website info
-    _update_website_metadata(bookmark)
     # Set currently logged in user as owner
     bookmark.owner = current_user
     # Set dates
     bookmark.date_added = timezone.now()
     bookmark.date_modified = timezone.now()
     bookmark.save()
+    # Update website info
+    _update_website_metadata(bookmark.id)
     # Update tag list
     _update_bookmark_tags(bookmark, tag_string, current_user)
     bookmark.save()
@@ -47,7 +48,7 @@ def update_bookmark(bookmark: Bookmark, tag_string, current_user: User):
         # Update web archive snapshot, if URL changed
         tasks.create_web_archive_snapshot(current_user, bookmark, True)
         # Only update website metadata if URL changed
-        _update_website_metadata(bookmark)
+        _update_website_metadata(bookmark.id)
 
     return bookmark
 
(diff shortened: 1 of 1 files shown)


tecoholic:
@sissbruecker Happy holidays. :christmas_tree:  Kindly take a look at this PR when you get some time after the vacation.

sissbruecker:
Thanks, it's an interesting idea, however there are some issues with it:
- When using the add bookmark form, the user gets redirected to the bookmark list after saving. Here they will only see the URL for the newly added bookmark, making it look like the website scaping is broken.
- Similarly, when having a large backlog of background tasks (e.g. after bookmark import), it could take a long time until the scaping eventually happens
- Background tasks can be disabled completely, which [some users might have done](https://github.com/sissbruecker/linkding/issues/157#issuecomment-1007706160):
https://github.com/sissbruecker/linkding/blob/2e3010d522003a13c39abd72cca3ce6c75c96ab5/bootstrap.sh#L19-L22

If anything, this should be an optional behavior, and there would need to be something in the bookmark list that indicates that the bookmark has not been scaped yet.

Personally, I'm currently more interested in making the scraping faster:
- https://github.com/sissbruecker/linkding/pull/354 - This change has already been merged, but is not released yet. Rather than loading the full HTML document, the scaper now stops loading after detecting the end of the head tag or after loading a fixed amount of data. This should help quite a bit.
- Another option I have been considering is skipping scraping on save. Currently when using the in-app bookmark form, the scaping actually runs twice: Once when opening the form, to show previews for website title and description, and a second time when saving the bookmark. If the form already has scraped the title and description, then the server shouldn't need to scape again, and instead the values could just be provided by the form. This would require some additional work to do the same for the extension.

tecoholic:
@sissbruecker Thank you for explaining the situation more clearly. I did miss the part in the documentation which says, the background tasks can be turned off. Regarding the 2 alternates you have mentioned here:

1. Limiting the document size of the website scrapper - I think this is such a great improvement and will definitely help bring down those times. :clap:
2. Preventing double scraping - Can't we do something simpler like using a `@functools.cache` to achieve the same? I will investigate this and open a separate PR if that is a possible solution.

I am closing this PR for now. :)

sissbruecker:
Thanks, temporarily caching the meta data might actually be a much easier alternative.

PR #383, opened 2022-12-25
Trim website metadata title and description

luca1197:
Some sites seem to return metadata titles / descriptions with trailing or leading whitespaces, which should be unwanted in most if not all cases. This PR adds a [`.trim`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim) method call to the fetched values, which will remove these whitespaces.

The whitespaces are not shown in the bookmark list after it is saved, but they are being stored, so they are also still there when editing the bookmark.

One example would be [Humble Bundle](https://www.humblebundle.com/) pages, e.g. this one: https://www.humblebundle.com/software/unreal-engine-ultimate-game-dev-assets-software

image

### DIFF
--- a/bookmarks/services/website_loader.py
+++ b/bookmarks/services/website_loader.py
@@ -29,9 +29,9 @@ def load_website_metadata(url: str):
         page_text = load_page(url)
         soup = BeautifulSoup(page_text, 'html.parser')
 
-        title = soup.title.string if soup.title is not None else None
+        title = soup.title.string.strip() if soup.title is not None else None
         description_tag = soup.find('meta', attrs={'name': 'description'})
-        description = description_tag['content'] if description_tag is not None else None
+        description = description_tag['content'].strip() if description_tag is not None else None
     finally:
         return WebsiteMetadata(url=url, title=title, description=description)
 
--- a/bookmarks/services/website_loader.py
+++ b/bookmarks/services/website_loader.py
@@ -31,7 +31,7 @@ def load_website_metadata(url: str):
 
         title = soup.title.string.strip() if soup.title is not None else None
         description_tag = soup.find('meta', attrs={'name': 'description'})
-        description = description_tag['content'].strip() if description_tag is not None else None
+        description = description = description_tag['content'].strip() if description_tag and description_tag['content'] else None
     finally:
         return WebsiteMetadata(url=url, title=title, description=description)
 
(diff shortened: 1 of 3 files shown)


sissbruecker:
Thanks, however the change here only affects the visual appearance of the placeholder, while the data returned from the scaper, and the data saved to the database, still contains the whitescape. It would make more sense to apply this fix to the scaper:
https://github.com/sissbruecker/linkding/blob/3abdd924303f557ff2bd342aeb96f5905649c4c1/bookmarks/services/website_loader.py#L25-L36

luca1197:
Thank you for the feedback, I did not expect it would also fetch the data in the backend again. I implemented the trimming in the scraper (thank you for the snippet!) and tested it. Seems to work great.

bookmarks/services/website_loader.py
```diff
@@ -29,9 +29,9 @@ def load_website_metadata(url: str):
         page_text = load_page(url)
         soup = BeautifulSoup(page_text, 'html.parser')
 
-        title = soup.title.string if soup.title is not None else None
+        title = soup.title.string.strip() if soup.title is not None else None
         description_tag = soup.find('meta', attrs={'name': 'description'})
-        description = description_tag['content'] if description_tag is not None else None
+        description = description_tag['content'].strip() if description_tag is not None else None
```
sissbruecker, line 34:
```suggestion
        description = description_tag['content'].strip() if description_tag and description_tag['content'] else None
```
I think this could result in a null reference if the `content` attribute does not exist.

bookmarks/templates/bookmarks/form.html
```diff
@@ -139,7 +139,7 @@
 
       function updatePlaceholder(input, value) {
         if (value) {
-          input.setAttribute('placeholder', value);
+          input.setAttribute('placeholder', value.trim());
```
sissbruecker, line 142: Doesn't look like this change is still necessary?

sissbruecker:
Added a test for the new trim behavior.

bookmarks/services/website_loader.py
sissbruecker: Done

bookmarks/templates/bookmarks/form.html
sissbruecker: Done

PR #387, opened 2023-01-06
Android HTTP shortcuts v3

kzshantonu:
This simplifies tagging and also filters URL's shared from apps. Useful for apps that don't share clean URLs

Co-Authored-By: Roland Meyer <2445563+Waboodoo@users.noreply.github.com>


sissbruecker:
Thanks @kzshantonu. As I would still not be able to maintain this myself, I was wondering if it would make more sense to move this to its own Github repo, and then link to that from the how-to docs page? Then you could improve this whenever you want, and people looking to collaborate on it could do so in your repo. What do you think?

I'll still merge this for now.

PR #392, opened 2023-01-13
Add Health Check endpoint 

### LINKED ISSUE #375: Add Health Check Endpoint for Container
It would be better if we have healthcheck endpoint if container goes unhealthy or down just like this one:

https://miniflux.app/docs/api.html#endpoint-healthcheck

mckennajones:
Adds a simple health check endpoint as mentioned in #375. The endpoint at `/health` simply returns the version of the app and a health status. It will return 'healthy` unless it is unable to connect to the DB. Could easily add other checks as deemed appropriate. 

Also added the health check to the Dockerfile so it will report it's health status. Example:

```
(venv) ➜  ~/temp/linkding git:(feature/health-endpoint) docker-compose up -d
[+] Running 1/1
 ⠿ Container linkding  Started                                                                                                                                         0.3s
(venv) ➜  ~/temp/linkding git:(feature/health-endpoint) docker ps
CONTAINER ID   IMAGE             COMMAND            CREATED          STATUS                            PORTS                    NAMES
19cd852b2d49   linkding-test:1   "./bootstrap.sh"   39 seconds ago   Up 3 seconds (health: starting)   0.0.0.0:9090->9090/tcp   linkding
(venv) ➜  ~/temp/linkding git:(feature/health-endpoint) docker ps
CONTAINER ID   IMAGE             COMMAND            CREATED          STATUS                             PORTS                    NAMES
19cd852b2d49   linkding-test:1   "./bootstrap.sh"   51 seconds ago   Up 15 seconds (health: starting)   0.0.0.0:9090->9090/tcp   linkding
(venv) ➜  ~/temp/linkding git:(feature/health-endpoint) docker ps
CONTAINER ID   IMAGE             COMMAND            CREATED              STATUS                    PORTS                    NAMES
19cd852b2d49   linkding-test:1   "./bootstrap.sh"   About a minute ago   Up 38 seconds (healthy)   0.0.0.0:9090->9090/tcp   linkding
(venv) ➜  ~/temp/linkding git:(feature/health-endpoint) rm -rf ./data

Wait 1 min 30s for 3 retries to occur.

(venv) ➜  ~/temp/linkding git:(feature/health-endpoint) docker ps
CONTAINER ID   IMAGE             COMMAND            CREATED         STATUS                     PORTS                    NAMES
19cd852b2d49   linkding-test:1   "./bootstrap.sh"   2 minutes ago   Up 2 minutes (unhealthy)   0.0.0.0:9090->9090/tcp   linkding

```

Btw, I'm not super familiar with django. Let me know if a middleware is not the appropriate way to add an endpoint like this. 

Closes #375

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,6 +53,6 @@ RUN ["chmod", "g+w", "."]
 RUN ["chmod", "+x", "./bootstrap.sh"]
 
 HEALTHCHECK --interval=30s --retries=3 --timeout=1s \
-CMD curl -f http://localhost:9090/health || exit 1
+CMD curl -f http://localhost:${LD_SERVER_PORT:-9090}/health || exit 1
 
 CMD ["./bootstrap.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ RUN mkdir /opt/venv && \
 
 
 FROM python:3.10.6-slim-buster as final
-RUN apt-get update && apt-get -y install mime-support libpq-dev
+RUN apt-get update && apt-get -y install mime-support libpq-dev curl
 WORKDIR /etc/linkding
 # copy prod dependencies
 COPY --from=prod-deps /opt/venv /opt/venv
@@ -51,4 +51,8 @@ ENV PATH /opt/venv/bin:$PATH
 RUN ["chmod", "g+w", "."]
 # Run bootstrap logic
 RUN ["chmod", "+x", "./bootstrap.sh"]
+
+HEALTHCHECK --interval=30s --retries=3 --timeout=1s \
+CMD curl -f http://localhost:9090/health || exit 1
+
 CMD ["./bootstrap.sh"]
--- a/bookmarks/middlewares.py
+++ b/bookmarks/middlewares.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.contrib.auth.middleware import RemoteUserMiddleware
 
+
 class CustomRemoteUserMiddleware(RemoteUserMiddleware):
     header = settings.LD_AUTH_PROXY_USERNAME_HEADER
--- a/bookmarks/middlewares.py
+++ b/bookmarks/middlewares.py
@@ -1,6 +1,30 @@
 from django.conf import settings
 from django.contrib.auth.middleware import RemoteUserMiddleware
-
+from django.http import JsonResponse
+from django.db import connections
+from bookmarks.views.settings import app_version
 
 class CustomRemoteUserMiddleware(RemoteUserMiddleware):
     header = settings.LD_AUTH_PROXY_USERNAME_HEADER
+
+class HealthCheckMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if request.path == '/health':
+            code = 200
+            response = {
+                'version': app_version,
+                'status': 'healthy'
+                }
+
+            try:
+                connections['default'].ensure_connection()
+            except Exception as e:
+                response['status'] = 'unhealthy'
+                code = 500
+
+            return JsonResponse(response, status=code)
+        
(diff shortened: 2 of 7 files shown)


bookmarks/middlewares.py
```diff
@@ -1,6 +1,30 @@
...
 
 class CustomRemoteUserMiddleware(RemoteUserMiddleware):
     header = settings.LD_AUTH_PROXY_USERNAME_HEADER
+
+class HealthCheckMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if request.path == '/health':
```
sissbruecker, line 15: It looks like this isn't called when the database is unable to connect. In `siteroot/settings/base.py` I modified the SQLite path to something invalid, then started the app. Accessing `localhost:8000/health` returns `Not found`. It works if the connection params are correct.

siteroot/settings/base.py
```diff
@@ -46,6 +46,7 @@
 ]
 
 MIDDLEWARE = [
+    'bookmarks.middlewares.HealthCheckMiddleware',
```
sissbruecker, line 49: A Django view would be a better option, but let's resolve the previous comment first.

bookmarks/middlewares.py
sissbruecker: Seems that is only the case for the Django dev server, it works fine when running the app through uwsgi. Feel free to to disregard then.

siteroot/settings/base.py
sissbruecker: To convert this into a view:
- add a new `health.py` file to `bookmarks/views`
- add a function there, implementation should be everything within the `if request.path == '/health':` block, the condition itself is not needed
- register the view function in `bookmarks/urls.py`

Also would be great to have at tests for the view (healthy / unhealthy cases).

bookmarks/middlewares.py
mckennajones: In my experience the entire app doesn't seem to start unless the DB connection is valid. Are you seeing the same?

siteroot/settings/base.py
mckennajones: Thanks @sissbruecker this is super helpful! I went ahead and converted it to a view. 

I also added a simple test for the healthy case. I wasn't sure how I could simulate the DB being down in a test. Any ideas?

bookmarks/middlewares.py
sissbruecker: Yeah, something like that. Should be fine to ignore in the dev environment.

siteroot/settings/base.py
sissbruecker: Added a test for the unhealthy scenario.

Dockerfile
```diff
@@ -51,4 +51,8 @@ ENV PATH /opt/venv/bin:$PATH
 RUN ["chmod", "g+w", "."]
 # Run bootstrap logic
 RUN ["chmod", "+x", "./bootstrap.sh"]
+
+HEALTHCHECK --interval=30s --retries=3 --timeout=1s \
+CMD curl -f http://localhost:9090/health || exit 1
```
sissbruecker, line 56: linkding has an option for changing the server port:
https://github.com/sissbruecker/linkding/blob/43d52642a69625f0706eae9e2ae7f61f20c3a24a/docs/Options.md#L65-L69

I think we need to change this so that the env variable is respected.

mckennajones: Aha good call, I missed that env var. Updated.

mckennajones:
Thanks for the commits @sissbruecker 🙂. Should be closer now. 

I also saw this older issue #272 and PR #273 that might be able to be closed out with the addition of this endpoint. Unless there is a use-case for an endpoint that solely returns the version.

sissbruecker:
Thanks, looks good now 👍

Thanks for the heads up, yes I think the health check endpoint should suffice for retrieving the version for now.

PR #393, opened 2023-01-15
Add opensearch to allow add search in webrowser

tomamplius:
Sorry i have check lot of time to understand namespace issue but i need help

error is bellow 
image
but if i replace opensearch by  bookmarks:opensearch in django-opensearch template, it's work

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,3 +29,4 @@ sqlparse==0.4.2
 typing-extensions==3.10.0.0
 urllib3==1.26.11
 waybackpy==3.0.6
+opensearch-py>=2.0.1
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -29,3 +29,4 @@ sqlparse==0.4.2
 typing-extensions==3.10.0.0
 urllib3==1.26.11
 waybackpy==3.0.6
+opensearch-py>=2.0.1
```
sissbruecker, line 32: I think you meant to add `django-opensearch`? `opensearch-py` seems like something different altogether: https://pypi.org/project/opensearch-py/

Besides, I don't think it's worth adding a new dependency for the little functionality that it provides. All that is needed is:
- a new link tag in the HTML
- a view that returns mostly static XML, with the exception of the dynamic search URL
- some tests to verify that link tag and XML are rendered correctly

tomamplius: yes i come back to my first implementation

PR #395, opened 2023-01-17
Disable autocapitalization for tag input form

joshdick:
I use lowercase names for all of my tags, so I find it annoying that the first tag I enter defaults to being capitalized (at least on mobile on iOS), and figured I'd open this PR to address it.

This change may be controversial [if others actually prefer this behavior](https://xkcd.com/1172/), but I feel that the current behavior of autocapitalizing only the first tag entered (as if inputting a sentence) is confusing at best.


joshdick:
@sissbruecker Do you have any comments/thoughts on this PR, or is there anything I can do to help move it along?

sissbruecker:
It seems reasonable, but I haven't really made up my mind yet.

PR #398, opened 2023-01-19
Allow to log real client ip in logs when using a reverse proxy

fmenabe:
Add a new *LD_LOG_X_FORWARDED_FOR* environment variable allowing to manage 
uWSGI [log-x-forwarded-for](https://uwsgi-docs.readthedocs.io/en/latest/Options.html?#log-x-forwarded-for) parameter. It allows to use the ip from *X-Forwarded-For* header instead of *REMOTE_ADDR* in logs and thus to keep real client IP when using a reverse proxy.


Q3x7Dvh:
Is there anything specific that is preventing this option being merged? I manually applied this patch to my server running the latest version (1.17.2) and there were no conflicts. I am getting the real IP addresses in my log now, which is awesome!

PR #403, opened 2023-01-21
Add opensearch

tomamplius:
Hi @sissbruecker ,

You can find without plugin opensearch implementation for review


bookmarks/views/bookmarks.py
```diff
@@ -16,6 +16,8 @@
 
 _default_page_size = 30
 
+def opensearch(request):
+    return render(request, 'bookmarks/opensearch.xml')
```
sissbruecker, line 20: Let's add a test to verify the XML template is rendered correctly with the respective dynamic URLs.

bookmarks/templates/bookmarks/opensearch.xml
```diff
@@ -0,0 +1,13 @@
...
+<Url type="text/html" method="GET" template="{{request.scheme}}://{{request.get_host}}/bookmarks?q={searchTerms}&amp;client=opensearch" />
```
sissbruecker, line 9: The absolute URL can be generated by Django, see the settings view for reference, which generates an absolute URL for the bookmark form:
https://github.com/sissbruecker/linkding/blob/4cb39fae99af0769f7a2bc8ece181085d8259b96/bookmarks/views/settings.py#L78

Preferrably this would happen in the view logic, not in the template itself.

bookmarks/templates/bookmarks/opensearch.xml
```diff
@@ -0,0 +1,13 @@
...
+<Url type="application/x-suggestions+json" method="GET" template="{{request.scheme}}://{{request.get_host}}/api/bookmarks/?limit=15&amp;offset=0&amp;q={searchTerms}" />
+<Query role="example" searchTerms="Linkding" />
+<moz:SearchForm>{{request.get_host}}</moz:SearchForm>
```
sissbruecker, line 12: The absolute URL can be generated by Django, see the settings view for reference, which generates an absolute URL for the bookmark form:
https://github.com/sissbruecker/linkding/blob/4cb39fae99af0769f7a2bc8ece181085d8259b96/bookmarks/views/settings.py#L78

Preferrably this would happen in the view logic, not in the template itself.

sissbruecker (commented):
I'm mostly fine with adding basic support for using linkding as a search engine. I'm probably not interested in doing much more with it, as the existing browser extension does basically the same, and has some additional benefits such as being able to suggest URLs that can be immediately opened.

sissbruecker:
Closing as stale

PR #407, opened 2023-01-23
Make health check in Dockerfile honor context path setting

mrex:
When using the context path setting the health check defined in Dockerfile fails, reporting the container state as unhealthy. That's because the path to the health endpoint is hardcoded to `/health` in Dockerfile, while the actual endpoint is provided under `${LD_CONTEXT_PATH}/health`.

This patch changes the path in Dockerfile to honor the value of LD_CONTEXT_PATH.


sissbruecker:
Thanks!

PR #411, opened 2023-01-27
Add postgres samples and instruction

kzshantonu:


sissbruecker:
Thanks for your effort, however I don't think I want to go ahead with this. The reason is that I don't want to end up maintaining and providing support for multiple setups. As mentioned in the other PR, I'd encourage people to create their own repos, and then link them from the community section.

Issue #43, opened 2020-12-28
Import should scrape title and description from URL


sissbruecker:
Currently bookmarks automatically use the title + description from the website when added manually.

The same should also work when importing bookmarks. This would be easy to add, but might slow down the import immensely due to the high number or requests that need to be executed. Would have to check if that is acceptable or if this can be solved through async jobs/tasks or similar.

kamikazechaser:
Could this be manually triggered using a script? or something similar to the /archive endpoint and a force an update?

PR #414, opened 2023-01-30
feat: add refresh description button to scrape the bookmark descriptions

leangseu:
Problem: after import. I didn't have any description. I would like the application to be able to do auto scrape of the description.

image

Screenshot 2023-01-29 at 9 27 38 PM

Screenshot 2023-01-29 at 9 27 44 PM

### DIFF
--- a/bookmarks/services/tasks.py
+++ b/bookmarks/services/tasks.py
@@ -11,7 +11,7 @@ from waybackpy.exceptions import WaybackError, TooManyRequestsError, NoCDXRecord
 import bookmarks.services.wayback
 from bookmarks.models import Bookmark, UserProfile
 from bookmarks.services import favicon_loader
-from bookmarks.services.website_loader import DEFAULT_USER_AGENT
+from bookmarks.services.website_loader import DEFAULT_USER_AGENT, load_website_metadata
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


bookmarks/services/tasks.py
```diff
@@ -161,6 +161,40 @@ def schedule_refresh_favicons(user: User):
...
+        bookmark = Bookmark.objects.get(id=bookmark_id)
+    except Bookmark.DoesNotExist:
+        return
+
+    logger.info(f'Load description for bookmark. url={bookmark.url}')
+
+    description = load_website_metadata(bookmark.url).description
+
+    if description is not None and description != bookmark.description:
+        bookmark.description = metadata.description
```
tchapi, line 193: 🌱  `metadata` is not defined here. You can just do `bookmark.description = description` (same for the log afterwards)

tchapi:
Hi @sissbruecker 👋🏼 
Any chance this could be merged at some point? This is a very handy feature!

Thanks!

kljohann:
Maybe add an option to only populate the description field if its empty instead of replacing existing descriptions of all bookmarks?

sissbruecker:
I'm not sure about the utility of having a button that refreshes the scraped descriptions of all bookmarks. That could be pretty annoying if you are used to searching for something that is in the description of a bookmark, and then suddenly it's gone.

The only case where you wouldn't have a description is if you import a bookmarks HTML file, and some bookmarks in that file don't have a description already. In that case it would be more useful to run this background task automatically after the import, as suggested in https://github.com/sissbruecker/linkding/issues/43.

Another thing to note is that there is a distinction between values (title + description) that have been scraped from the website (`website_title` and `website_description` fields) , and values that have been entered manually (`title` and `description` fields). The app will always prefer to show the manually entered values, and only show the scraped values if there are no manual ones:
https://github.com/sissbruecker/linkding/blob/5049ff14cff1c5442e0765bf44286b68f09051d6/bookmarks/models.py#L66-L77

This PR currently overrides the manually entered description with the one scraped from the website, which should not happen.

sissbruecker:
Closing as stale

Laephis:
Any chance this could be reopened and looked at again? I recently installed linkding and imported 900 bookmarks but nearly all of them have the URL as the title.  I see that I can go back and edit each one of them manually by clicking "refresh from website", but with so many that's just not feasible.  Having an automated way to do it would be very helpful.  Or maybe an admin action option in Bookmarks to bulk edit/refresh the Resolved Title similar to how we can perform some other tasks?

Issue #380, opened 2022-12-22
[Feature Request] Add ArchiveBox as an archival endpoint


ctag:
[ArchiveBox](https://archivebox.io/) is a self-hosted web archival tool that stores pages in a variety of formats. While it appears there [isn't a stable REST API yet](https://github.com/ArchiveBox/ArchiveBox/issues/496), there is a pretty simple CLI that should be workable with linkding. 

Would a pull request for adding Archivebox support be given consideration?

sissbruecker:
Integrating this might be an option, but I'd prefer using a REST API rather than including a CLI tool into the Docker image, and then trying to drive that from the Django app. 

Apart from that the expectations should be clarified. Creating the snapshot seems pretty clear. What do you expect to happen **after** the snapshot was created?
- Should the snapshot link surface in the bookmark list?
- Should this be an alternative to the Wayback machine integration, and replace the snapshot URL created by it?

An alternative could be to expose the `web_archive_snapshot_url` field in linkding's REST API, which would be fairly trivial to do, and then write a 3rd party app that:
- Checks linkding for bookmarks that don't have a snapshot URL yet
- Create an archive box snapshot for bookmarks that don't have one yet
- Update the bookmark's snapshot URL in linkding

pratio:
@sissbruecker It would be great if the snapshot link would replace the internet archive link. It would be enough if we can manually trigger it for the bookmarks that haven't been archived from the admin panel.

webysther:
REST API is arrived: https://github.com/ArchiveBox/ArchiveBox/pull/1397#issuecomment-2076913352

huyz:
How would this be different from having ArchiveBox pull an RSS feed from LinkDing?

webysther:
> How would this be different from having ArchiveBox pull an RSS feed from LinkDing?

I have no ideia, how do that?

Kos:
I've did my first Archivebox backup of my Linkding feed yesterday. It was very easy: I grabbed the RSS link and added it to Archivebox UI with depth=1 and my archive methods of choice (wget + mercury + title + favicon).

I understand that Archivebox also [comes with a scheduler](https://github.com/ArchiveBox/ArchiveBox/wiki/Scheduled-Archiving) that can refresh the backups regularly, so once my Linkding RSS is registered, I expect that it will keep my archive updated.

(This is good enough for me.)

[...]

PR #434, opened 2023-02-24
feat: Archived Bookmarks RSS Feed

mayanez:
Hi!

First off, this is an awesome project. The codebase is super clean and simple. I love it! ❤️ 

I was originally looking for a way to archive to Archivebox and came across this issue https://github.com/sissbruecker/linkding/issues/380. I took a slightly different approach to the suggested to achieve part of this feature request.
 
Since Archivebox has support for parsing RSS, I've instead just exposed the archived bookmarks through that. All that is required to close the loop is schedule ArchiveBox to read the feed at the user's specified rate. For example, I have mine set to every 10 minutes. This also removes the need for the scheduling workload to be handled by linkding.

Granted, while the `Archive` link takes you to `archive.org` for now, as suggested in the issue above that can be exposed and updated in the settings/API. I'd be happy to work on this feature in the near future. For now with just the RSS feed it's sufficient to add the bookmarks URL to the Archivebox URL to get the actual snapshot.

### DIFF
--- a/bookmarks/feeds.py
+++ b/bookmarks/feeds.py
@@ -18,7 +18,7 @@ class BaseBookmarksFeed(Feed):
     def get_object(self, request, feed_key: str):
         feed_token = FeedToken.objects.get(key__exact=feed_key)
         query_string = request.GET.get('q')
-        query_set = queries.query_bookmarks(feed_token.user, query_string)
+        query_set = queries._base_bookmarks_query(feed_token.user, query_string)
         return FeedContext(feed_token, query_set)
 
     def item_title(self, item: Bookmark):
@@ -42,7 +42,7 @@ class AllBookmarksFeed(BaseBookmarksFeed):
         return reverse('bookmarks:feeds.all', args=[context.feed_token.key])
 
     def items(self, context: FeedContext):
-        return context.query_set
+        return context.query_set.filter(is_archived=False)
 
 
 class UnreadBookmarksFeed(BaseBookmarksFeed):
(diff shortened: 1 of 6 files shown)


sissbruecker:
Hey, it looks like there is some confusion around the archive function in linkding. Archiving or unarchiving a bookmark is a function that you have to trigger manually and does not relate to creating an archive of the website itself. The function was introduced to get bookmarks that are not relevant anymore out of the way, but without actually deleting them. So the only property of archived bookmarks is that they don't show up in regular searches, or the RSS feed, but you can still find them in the archived bookmarks section.

The web archive integration in comparison just creates snapshots of all bookmarks that have been added. So if you want to replicate that with ArchiveBox, then use the all bookmarks RSS feed.

#380 seems still relevant as for an integration I would expect some way to link from linkding to ArchiveBox. If you have some ideas on how to achieve that, I would suggest to add some proposal to the issue first. Also it would be nice if you could share your approach of using the RSS feed in the issue so that others become aware of it.

I'll close this as I don't see a lot of value for providing an RSS feed for archived bookmarks.

mayanez:
I see. I agree my usage for archiving is slightly different in that I do not want to archive all bookmarks.

I was under the impression "Archived" bookmarks were simply a category no different than say "Unread". Thus, I was using the Archived bookmarks to mark those I want to send to ArchiveBox. It seemed more appropriate to expose that "category" via the RSS feed rather than using tags and filtering the RSS feed to only get those bookmarks I wanted to archive.

I think it is beneficial to expose the archived bookmarks in the same vein that "Unread" bookmarks are also exposed.

sissbruecker:
To clarify this further, how do you intend to use the linkding UI then? Some of your bookmarks will be in the active (default) view, while others will only be visible in the archived view. Search does not include both active and archived bookmarks, so you have to search in both views if you want to find something specific. The extension only searches for active bookmarks.

mayanez:
I had planned to use it with the separation as it is now. I haven't explored using the extension yet to be honest, but given that it only currently searches for active bookmarks I might have added a setting to toggle enabling searching through the archived bookmarks as well. I haven't yet looked to see if this is something you've previously considered or discounted.

However, if this usage of Archived bookmarks is not in scope then I will fall back to using tags to categorize and filter bookmarks to feed into ArchiveBox that way.

Issue #409, opened 2023-01-26
Search doesn't match against tags unless use #


adamshand:
Hello,

Currently, linkding searches don't include tags unless you explicitly include the #. 

For example, I just added a link with a tag `interviewing`, but if I search for `interviewing` it doesn't show up (it does show up if I search for `#interviewing`).

It would be very helpful for me if tags were searched along with all the other fields.

Even better than that would be if:

- searching for `interviewing` returns links which have interviewing in any field (including tags)
- searching for `#interview` only returns links which have the tag `interviewing`

This seems more explicit and obvious to me, by default searches return all matching results.  If you only want to restrict searches to only tags, use the # prefix.

Thanks again for linkding, I use it every day!

See also: https://github.com/Fivefold/linkding-injector/issues/8

Cheers,
Adam.

nickludlam:
I'd also love this as either a straight default, or an option I can turn on. I just had an issue where a keyword I searched for was in the tag, but not the title or description, so I thought I had lost the URL

mckennajones:
Opened a small PR to add support for this. Ended up being a one-liner.

mckennajones:
@sissbruecker has a better approach on his [feat/search_tags_without_hash](https://github.com/sissbruecker/linkding/compare/feat/search_tags_without_hash) branch, so closed my PR in favor of that.

PR #436, opened 2023-03-02
Update search to include tags even if not explicitly searching for tags

mckennajones:
Updates search to follow the behavior proposed in #409:
- Searching for a term returns results which contains the term in any field, including tags.
- Searching for a tag with `#` only returns those which are tagged with the tag. 

@sissbruecker let me know what you think about leaving this as the default or hiding it behind a setting

### DIFF
--- a/bookmarks/queries.py
+++ b/bookmarks/queries.py
@@ -41,6 +41,7 @@ def _base_bookmarks_query(user: Optional[User], query_string: str) -> QuerySet:
             | Q(website_title__icontains=term)
             | Q(website_description__icontains=term)
             | Q(url__icontains=term)
+            | Q(tags__name__icontains=term)
         )
 
     for tag_name in query['tag_names']:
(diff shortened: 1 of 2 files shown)


sissbruecker:
The approach here creates a left outer join, which means if a matching bookmark has more than one tag, you get duplicates in the search result.

I actually worked on this a bit last week, and ended up using a sub-select: https://github.com/sissbruecker/linkding/commit/7a10cbc4b392525cf44b8846e5c93240e70c92d0#diff-dbebb373746707e3e502fe8ee300bc90a8b3fbc0a24e76e42bed71453dbf6096.
I think it would be best to put this behind an option for now, maybe make it the default later. Was planning on finishing this in the near future.

mckennajones:
Aha I didn't catch that, thanks @sissbruecker. I'll abandon this PR then.

PR #446, opened 2023-03-22
Update README.md

amoscardino:
I've released an iOS app for linkding called LinkThing. This update adds a link to it under the Community section of the readme


sissbruecker:
Cool, thanks!

Can the app be used as URL share target? What I mean is, when you share a URL from Safari for example, can you share it to LinkThing so that you can customize the bookmark data before saving it to linkding?

amoscardino:
> Can the app be used as URL share target? What I mean is, when you share a URL from Safari for example, can you share it to LinkThing so that you can customize the bookmark data before saving it to linkding?

Not yet, though I really want to add that as a feature. I'm just not sure how to accomplish it with the framework I'm using.

Issue #382, opened 2022-12-24
 [Feature Request] API Enhancement for Fetching Bookmarks/Tags that have been Modified or Created after Date


jenisgrindstaff:
Would it be possible to add a feature to the api to only retrieve objects that have been updated or created after a certain date? This would prevent having to re-fetch all after the initial download.

sissbruecker:
That should be possible, each bookmark has a creation and update timestamp.

sissbruecker:
This is now possible for bookmarks. For tags it might be overkill, but I'm OK with someone contributing this.

PR #448, opened 2023-03-31
feat: add query date parameters

jpittiglio:
The purpose of this PR is to implement the functionality requested in #382 , specifically being able to query for only newly added or modified bookmarks (versus a full list every time).  This PR enables this through the use of additional URL parameters:

- `since_added`
- `until_added`
- `since_modified`
- `until_modified`

Note only the basic list bookmarks (`GET /api/bookmarks/`) has been updated, and not archived.  Happy to add this if needed or desired, but didn't want to go too far down a rabbit hole if this isn't in-line with desired implementation.

This PR only modifies the API; no front-end changes are made to support this.

### DIFF
--- a/bookmarks/api/routes.py
+++ b/bookmarks/api/routes.py
@@ -23,7 +23,11 @@ class BookmarkViewSet(viewsets.GenericViewSet,
         # For list action, use query set that applies search and tag projections
         if self.action == 'list':
             query_string = self.request.GET.get('q')
-            return queries.query_bookmarks(user, query_string)
+            query_date_keys = ['since_added', 'until_added', 'since_modified', 'until_modified']
+            query_dates = {
+                key: value for key, value in self.request.GET.items() if key in query_date_keys and value is not None
+            }
+            return queries.query_bookmarks(user, query_string, query_dates)
 
         # For single entity actions use default query set without projections
         return Bookmark.objects.all().filter(owner=user)
--- a/bookmarks/models.py
+++ b/bookmarks/models.py
@@ -123,6 +123,11 @@ class BookmarkFilters:
         self.query = request.GET.get('q') or ''
         self.user = request.GET.get('user') or ''
 
+        query_date_keys = ['since_added', 'until_added', 'since_modified', 'until_modified']
+        self.query_dates = {
+            key: value for key, value in request.GET.items() if key in query_date_keys and value is not None
+        }
+
 
 class UserProfile(models.Model):
     THEME_AUTO = 'auto'
--- a/bookmarks/queries.py
+++ b/bookmarks/queries.py
@@ -2,29 +2,32 @@ from typing import Optional
 
 from django.contrib.auth.models import User
 from django.db.models import Q, QuerySet
+from django.utils import timezone
 
 from bookmarks.models import Bookmark, Tag
 from bookmarks.utils import unique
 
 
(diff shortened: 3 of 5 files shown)


bookmarks/models.py
```diff
@@ -123,6 +123,11 @@ def __init__(self, request: WSGIRequest):
         self.query = request.GET.get('q') or ''
         self.user = request.GET.get('user') or ''
 
+        query_date_keys = ['since_added', 'until_added', 'since_modified', 'until_modified']
```
sissbruecker, line 126: It took my brain a few turns to make sense of `until_added`, in general I think it would be easier to understand if the words were reversed.

How about:
- `added_since`
- `added_before`
- `modified_since`
- `modified_before`

bookmarks/queries.py
```diff
@@ -59,6 +62,24 @@ def _base_bookmarks_query(user: Optional[User], query_string: str) -> QuerySet:
...
 
+    # Ensure dates are converted to correct `datetime` representation
+    dates = parse_query_dates(query_dates)
+
+    # Filter by dates
+    date_filters = {
+        'since_added': 'date_added__gte',
+        'until_added': 'date_added__lte',
+        'since_modified': 'date_modified__gte',
+        'until_modified': 'date_modified__lte'
```
sissbruecker, line 73: Have you thought about making one boundary exclusive? It's probably not a big deal in practice, but maybe you had some thoughts.

bookmarks/tests/test_bookmarks_api.py
```diff
@@ -60,6 +62,22 @@ def test_list_bookmarks_should_filter_by_query(self):
                             expected_status_code=status.HTTP_200_OK)
         self.assertBookmarkListEqual(response.data['results'], [self.bookmark1])
 
+    def test_list_bookmarks_should_filter_by_timestamp(self):
```
sissbruecker, line 65: Tests could be a bit more extensive to cover each param, and verify that each actually returns results - for example the `until_added` case below only checks for an empty result, which might also be a bug. It should be trivial to create a few bookmarks with added/modified dates for the test setup.

Ideally this set of extensive tests would be in `test_queries.py`, which contains tests for all the other query logic. Then have a simpler test case in this file to test integration with the API.

bookmarks/api/routes.py
```diff
@@ -23,7 +23,11 @@ def get_queryset(self):
         # For list action, use query set that applies search and tag projections
         if self.action == 'list':
             query_string = self.request.GET.get('q')
-            return queries.query_bookmarks(user, query_string)
+            query_date_keys = ['since_added', 'until_added', 'since_modified', 'until_modified']
+            query_dates = {
+                key: value for key, value in self.request.GET.items() if key in query_date_keys and value is not None
+            }
+            return queries.query_bookmarks(user, query_string, query_dates)
```
sissbruecker, line 30: Would make sense to reuse `BookmarkFilters` here.

Looking at the queries module I think long-term it would make sense to replace the individual filter params with the filters object itself, but that can be refactored later.

sissbruecker:
Closing as stale

mzehrer:
What is needed to integrate this? This would make the API really useful for integrating with other services.

sissbruecker:
Someone needs to address the review comments and open a new PR.

Issue #327, opened 2022-08-18
Comments for Links


tobru:
In addition to the "Description" field, it would be cool to have a new "Comments" field. I usually like to add a few notes or thoughts to a link, why I bookmarked a link for example, but leaving the description intact. For now, I'm using the description field, but feels semantically incorrect to do so, as I'm now mixing the description and my comments into the same field.

michaelx:
Agreed. An additional field for comments/notes with Markdown support would be helpful.

turkishmaid:
Supporting this enhancement request. Also the comments field should be well bigger than the current description. I use to add lots of info from the linked page in the description (for future quoting etc.), and the two lines turns out to be quite a Sehschlitz to the text in this case. This is even more annoying with iOS Safari which does not sow me the resize handle of the texture, so it's really only two lines.

catusax:
i have made a commit witch add a comment filed, see https://github.com/catusax/linkding/commit/5795cc5ab7085be379a3ad5e85a2ec48744b584c
But it affects database table and Web UI. I think it's a big decision, should i open a pull request? @sissbruecker

PR #472, opened 2023-05-20
Add notes to bookmarks

sissbruecker:
Allows adding additional notes to bookmarks, with Markdown support. Notes are hidden by default, and can be toggled to be visible per bookmark. A new setting allows to make notes permanently visible. Alternatively the keyboard shortcut <kbd>e</kbd> allows to temporarily show all notes.

Closes https://github.com/sissbruecker/linkding/issues/327


tobru:
Wohoo, thank you very much, this is greatly appreciated.

Issue #465, opened 2023-05-09
Improvements to favicons


Zottelchen:
I really like the favicon function, but here are some possible improvements I noticed:

1. Let the service be selected in the settings.
2. Offer an alternative to the _{url}_ replacer (➡e.g. `https://www.github.com`) - _{domain}_ (➡ `www.github.com`) would help to make it easily work with other services 
3. Don't assume the returned Icon is a png. It could be an .ico
4. Extend the documentation with alternative icon providers:
     * `https://icons.duckduckgo.com/ip3/{domain}.ico` - a privacy respecting alternative
     * `https://besticon-demo.herokuapp.com/icon?url={url}&size=32..64..128&formats=png` - flexible & selfhostable
     * `https://icon.horse/icon/{domain}`
     * `https://favicongrabber.com/api/grab/{domain}` - requires additional work
 5. If using Google by default, this seems to give me better results: `https://t1.gstatic.com/faviconV2?client=SOCIAL&type=FAVICON&fallback_opts=TYPE,SIZE,URL&url={url}&size=64`

sissbruecker:
Thanks, this seems useful. Offering some more choices for icon providers would be great. Making them selectable in the options is more of a nice to have for me.

Some clarification:
2. Could you elaborate a little bit? I'm not quite sure what you are suggesting, and I'm also not in the context anymore on how the current url placeholder works.
3. Probably doable by inspecting the response headers. Thankfully the extension is also stored in the database field, so it should be doable without breaking changes.
5. Better results how? Like what is better about it?

Zottelchen:
Sure :)
2. Basically right now the replacer is `{url}` - it includes the protocol (typically `https://`). Some icon providers just want the basedomain though without the protocol.
[...]

PR #504, opened 2023-08-15
Various improvements to favicons

sissbruecker:
This adds various improvements to how favicons work:
- Changes the default favicon provider to use a higher-resolution image to give better results for high DPI displays
- Adds a `{domain}` placeholder for custom favicon providers that require only the domain / host name without the URL scheme
- Automatically tries to detect the correct file extension for the loaded favicon
- Adds DuckDuckGo as an alternative favicon provider to the documentation

Closes https://github.com/sissbruecker/linkding/issues/465

### DIFF
--- a/bookmarks/services/favicon_loader.py
+++ b/bookmarks/services/favicon_loader.py
@@ -1,3 +1,4 @@
+import logging
 import os.path
 import re
 import shutil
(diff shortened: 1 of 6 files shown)


AprilNEA:
Good job, my icon service hasn't worked for a long time now.

PR #512, opened 2023-08-24
Search suggestions now follows link_target user's setting

### LINKED ISSUE #511: Search suggestions open in new window regardless of setting
While the open in "new page/same page" setting works fine for all the bookmarks listed on the page, clicking any bookmark from the suggestion box always opens the url in a new page, regardless of the user's setting.

rootforbid:
Fixes #511

### DIFF
--- a/bookmarks/api/routes.py
+++ b/bookmarks/api/routes.py
@@ -5,8 +5,8 @@ from rest_framework.response import Response
 from rest_framework.routers import DefaultRouter
 
 from bookmarks import queries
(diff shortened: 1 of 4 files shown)


sissbruecker:
Thanks, but rather than initiating another request to fetch the user profile it is easier, and more idiomatic, to pass the setting into the component when rendering the template on the server. Opened https://github.com/sissbruecker/linkding/pull/513 to address that and closing in favor of that.

PR #513, opened 2023-08-24
Make search autocomplete respect link target setting

### LINKED ISSUE #511: Search suggestions open in new window regardless of setting

sissbruecker:
Fixes #511


rootforbid:
Yes, it's a much better solution, my Django is a bit rusty...
Thanks for taking care of the fix :+1:

Issue #484, opened 2023-06-02
Expanded support for shared flag


gingerbeardman:
If this is in the roadmap and I have missed it, apologies.

Soon after issue #204 was closed initial support for shared tag was added to Linkding.

Shared can be set and queried using the API, but full support is not yet reached.

Observations:
- no front end way to set shared flag when adding a bookmark (in the same way as "Mark as unread" works)
- shared status is not displayed in the list view
- import strips private flag rather than mapping it to shared
- exporting sets private=0/false even if bookmark has shared flag set as shared=true

gingerbeardman:
I have a bit of time coming up, I could get started on any import/export changes? any thoughts?

gingerbeardman:
Another weekend approaches, shall I begin work on this?

sissbruecker:
Some feedback on this:
- You need to enable bookmark sharing in the settings in order for a checkbox in the bookmark form to show up
- Shared status is indeed not shown in the list. I'm being careful with adding more stuff to bookmark items, in order to keep the list readable. An opinionated choice, but you're the first to want this, so maybe not very relevant.

Regarding mapping shared bookmarks to the private flag, I'm not convinced that this is the right approach, as these have different meanings:
- At least in Pinboard, a non-private bookmark is a public one, which means everyone, even anonymous users can see it.
- In Linkding, a shared bookmark is only shared with other authenticated users.

That results in two issues:
- Users, especially existing ones, might not expect that importing a non-private bookmark now suddenly shares it with everyone else in their installation
- If support for public bookmarks ever gets added, then the private flag is already mapped to something else

gingerbeardman:
Thanks for your thoughts.

>  but you're the first to want this

I think it would be easy to argue that the users in issue #204 also want it?

> Users, especially existing ones, might not expect that importing a non-private bookmark now suddenly shares it with everyone else in their installation

I can't speak for the number of installs that have multiple users, but this scenario would be easy to cope with by defaulting the setting to off (no change to behaviour, just like the one to enable bookmark sharing). 

> If support for public bookmarks ever gets added, then the private flag is already mapped to something else

My first thought is that the a private bool (0/1) could be transformed into an enum (0/1/2) for an easy migration to private/localshare/publicshare. Just an idea.

sissbruecker:
[...]

PR #515, opened 2023-08-24
Display shared state in bookmark list

sissbruecker:
Adds a button to the bookmark actions that shows the shared state of the bookmark. Clicking the button brings up a confirmation to unshare the bookmark. For consistency, the mark as read button also has been changed to instead show the state ("Unread"), rather than the action. Clicking that button also brings up a confirmation to mark the bookmark as read.

Bildschirmfoto 2023-08-24 um 18 58 59

Closes https://github.com/sissbruecker/linkding/issues/484


sissbruecker:
Still want to experiment with some other stuff before the next release. This PR also adds the `shared` class to bookmark list items if you want to tinker with that.

PR #547, opened 2023-10-06
Update supervisord.conf

Strubbl:
Move background_tasks.log to data/ directory, so that it can easily be read via the docker-shared volume.

I do not know if this makes sense to you. For me this helped in getting to see, that the background task from my import kept running. Feel free to decline this PR if not needed.

### DIFF
--- a/supervisord.conf
+++ b/supervisord.conf
@@ -5,7 +5,7 @@ loglevel=info
 [program:jobs]
 user=www-data
 command=sh background-tasks-wrapper.sh
-stdout_logfile=background_tasks.log
+stdout_logfile=data/background_tasks.log
 stdout_logfile_maxbytes=10MB
 stdout_logfile_backups=5
 redirect_stderr=true


sissbruecker:
Not sure if that is best practice or not. Can you provide some examples of other apps that handle logs like this? 

Alternatively, the logs could be put into a different folder like `/etc/linkding/logs` (some Linux user would scream right now), and then you could mount that in addition to the data folder.

Strubbl:
One could move linkding in the container to e.g. /app to not let the Linux users scream :D

Archivebox has a data directory and in there is a directory called `logs`. You can only see this when running the container:
https://github.com/ArchiveBox/ArchiveBox/blob/dev/docker-compose.yml
I have no other example at the moment.

Another alternative could be to redirect the background log job, so that it can be read via `docker-compose logs`. So stdout? Or have a second container only for background jobs?

sissbruecker:
For now I don't want to dilute the data folder with logs and only keep application data in there. Redirecting to stdout would be an option, but could be pretty noisy. I think I'd prefer using a separate folder for now.

sissbruecker:
Closing as stale, feel free to open a new PR if you want to continue.

PR #570, opened 2023-11-05
Add Alpine based Docker image (experimental)

sissbruecker:
Adds a Docker image that is based on Alpine Linux. This image is significantly smaller in size than the default Debian based image, resulting in a smaller download and less disk space required. According to Docker Hub, the Alpine based image is about 45 MB in compressed size compared to 130 MB for the Debian based image. There might be minor improvements in memory usage as well.

To use the new image, use the `latest-alpine` tag instead of `latest`.

> [!WARNING]
> The image is currently considered experimental in order to gather feedback and iron out any issues. Only use it if you are comfortable running experimental software or want to help out with testing. While there should be no issues with creating new installations, there might be issues when migrating existing installations. If you plan to migrate your existing installation, make sure to create proper [backups](https://github.com/sissbruecker/linkding/blob/master/docs/backup.md) first.


jonathan-s:
@sissbruecker Worth considering _not_ using alpine based on this blog post. https://pythonspeed.com/articles/alpine-docker-python/

Issue #454, opened 2023-04-17
Make backups easier


sissbruecker:
Everyone should do backups of their linkding database, however the current options are not straightforward:
- You can export manually through the UI - does not work automatically
- You can use an SQLite installation from the host system to export a backup - requires you to download SQLite first
- You can export everything through the REST API - requires advanced scripting, can only access bookmarks of single user, also need to restore backup somehow

This could be easier. One option I can think of is to utilize the SQLite binary that already exists in the Docker container to create the backup, and store the backup file in a mounted Docker volume. With that, users could just run a single command line prompt against the container to create the backup and store it on the host system. Then transfer the file to long-term storage. A more advanced version could be to create backups automatically on a schedule, so that you would only have to take care of the transfer part yourself.

virtadpt:
You can just copy a SQLite database somewhere, though.  It's a single file.

sissbruecker:
That gives you a good chance to backup a corrupted database in case the file is modified by the application while you are copying it. The database should be [locked](https://www.sqlite.org/backup.html) before copying the file, using the SQLite binary seems to be easiest way to do so.

bjzhush:
you can also use dropbox script to backup sqlite files automatically

lewis262626:
A backup to some form of object storage (whether S3, BlackBlaze B2, Minio) etc would be welcome

MaxRies:
@lewis262626 This is nicely implemented in the fly.io integration with lite-stream. Works like a charm.
https://github.com/fspoettel/linkding-on-fly
I'm sure you could use this for backing up your database. It would be handy to integrate this in the webapp, instead of configuring it via docker environment variables.

pettijohn:
Litestream for sqlite is a great option, I use it for this and a few other things I self-host. I'd also love a command line to export to HTML, maybe accept user as an argument or dump all with one file per user. Nothing beats plain text files for backups!

l0b0:
For reference, Pinboard provides https://pinboard.in/export/format:html/, https://pinboard.in/export/format:json/, and https://pinboard.in/export/format:xml/ links for logged-in users, so it's really easy to back up the content in a version-control-friendly fashion.

(I'm not a linkding user, but would probably be if this were implemented. This was the first feature I checked for after hearing about this project.)

PR #571, opened 2023-11-05
Add backup CLI command

sissbruecker:
Adds a CLI command that allows to create a backup copy of the SQLite database within the Docker container. The file can then be moved to the host system using `docker cp`.

Closes https://github.com/sissbruecker/linkding/issues/454

### DIFF
--- a/bookmarks/management/commands/backup.py
+++ b/bookmarks/management/commands/backup.py
@@ -1,0 +1,26 @@
+import sqlite3
+import os
+
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    help = "Creates a backup of the linkding database"
+
+    def add_arguments(self, parser):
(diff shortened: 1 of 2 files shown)


fs111:
I don't quite understand this feature. Are you not mounting a host directory inside the container to keep the db safe? I am currently simply copying the file in a cron job and that just works. Am I missing something?

sissbruecker:
Technically the copy approach is not safe: https://www.sqlite.org/howtocorrupt.html

Thus an explicit command for creating a backup.

fs111:
fair enough. My little linkding has so little traffic that I assume it is safe to do. There is no active transaction and if the file is corrupt I get a fresh one 4 hours later.

sissbruecker:
Yes, I would also assume there is an extremely low risk for the linkding use-case. But for an official recommendation for backups I'd rather go the safe route.

PR #581, opened 2023-11-27
Increase tag limit in tag autocomplete

### LINKED ISSUE #580: Tag completion doesn't work with newer tags
Hello. Is there a maximum number of tags possible? I'm asking because I noticed that I haven't been able to use tags completion when adding bookmarks if the tag in question was recently created. Completion with older tags works normally.

Since I have a huge number of tags (in fact it's more than 1400, with 2 different users) I imagine if I reached some limit, or something.

hypebeast:
Increased tag limit to 5000. It looks like that the tag limit of 1000 is reached by some users.

Fixes #580

### DIFF
--- a/bookmarks/frontend/components/TagAutocomplete.svelte
+++ b/bookmarks/frontend/components/TagAutocomplete.svelte
@@ -22,7 +22,7 @@
   async function init() {
     // For now we cache all tags on load as the template did before
     try {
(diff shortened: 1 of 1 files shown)


sissbruecker:
Thanks, I guess it's a reasonable workaround for now. 

Note that the browser extension duplicates this code, so that still has a limit of 1000.

PR #583, opened 2023-11-30
Added example command for restoring db.sqlite3

adamshand:


sissbruecker:
That's not really how it should work. The file should just be put into the host folder that is mounted as data folder into the Docker container. I see no need to provide a command to copy the file into the container, which is potentially confusing and just a convoluted way of putting it into the folder on the host system.

adamshand:
No worries. I was working with volumes instead of bind mounts, so it's not quite as straight forward to copy files directly into the filesystem.  The advantage of `using docker cp` from a documentation point of view is that it works for both bind mounts and volumes.

PR #602, opened 2024-01-14
Support Open Graph description (and introduce pytest for better dev ux)

jonathan-s:
As discussed in https://github.com/sissbruecker/linkding-extension/issues/59 you pointed out that the change should be made in this repo. This PR adds a test and the code to support that change. 

I also took the liberty of introducing pytest, because it's a much nicer experience for running tests when you're developing :).

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,6 @@
 </div>
 
 ## Overview
-- [Overview](#overview)
 - [Introduction](#introduction)
 - [Installation](#installation)
   - [Using Docker](#using-docker)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -7,18 +7,24 @@
 </div>
 
 ## Overview
```
sissbruecker, line 9: Let's revert changes to the TOC, I intentionally didn't include every item for a better overview.

bookmarks/tests/test_website_loader.py
```diff
@@ -94,3 +98,10 @@ def test_load_website_metadata_trims_title_and_description(self):
             metadata = website_loader.load_website_metadata('https://example.com')
             self.assertEqual('test title', metadata.title)
             self.assertEqual('test description', metadata.description)
+
+    def test_load_website_metadata_using_og_description(self):
```
sissbruecker, line 102: Ideally there would be a test that verifies that `name="description"` has precedence over the Open Graph metadata.

README.md
sissbruecker: Done

bookmarks/tests/test_website_loader.py
sissbruecker: Done

PR #607, opened 2024-01-19
Add title to link so you can see the entire title when hover

### LINKED ISSUE #606: Title is cut off if it is long > provide title attribute to link.

jonathan-s:
Fixes #606 

This is what it looks like with a custom hover over a link that has an overly long title. 

Screenshot 2024-01-19 at 16 04 59

Light theme.
Screenshot 2024-01-19 at 16 37 49

### DIFF
--- a/bookmarks/styles/bookmark-page.scss
+++ b/bookmarks/styles/bookmark-page.scss
@@ -107,6 +107,18 @@ ul.bookmark-list {
   padding: 0;
 }
 
+@keyframes appear {
+  0% {
+    opacity: 0;
+  }
+  90% {
+    opacity: 0;
+  }
+  100% {
+    opacity: 1;
+  }
+}
+
 /* Bookmarks */
 li[ld-bookmark-item] {
   position: relative;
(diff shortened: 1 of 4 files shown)


bookmarks/styles/bookmark-page.scss
```diff
@@ -122,6 +122,22 @@ li[ld-bookmark-item] {
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
+
+    &[data-title]:hover::after {
```
sissbruecker, line 126: While this is a neat trick, it's super annoying that this immediately pops up when you hover over a link. 

I would be more open to using the regular `title` attribute.

jonathan-s: @sissbruecker I hear you. I force-pushed a new commit which makes two amendments. 

- The pop-up will only show if the text is partially hidden.
- There is a ~300ms delay before the pop-up shows up.

sissbruecker: Thanks. Did some tweaks:
- Restrict tooltip max width to bookmark item so longer titles don't overflow
- Make tooltip show up on focus, for example when using keyboard navigation
- Refactored JS so that it reuses existing bookmark item behavior, doesn't need to duplicate title into a data attribute
- Refactored CSS to reuse Spectre CSS variables

jonathan-s:
Thank you!

PR #614, opened 2024-01-28
Activate CI for contributors and when pushing tags

jonathan-s:
I noticed that CI is not running when making a PR. It would make it easier if CI was actually running for people making a PR, both for yourself and for the contributor. 

There is the following setting as well. I think requiring approval for first time contributors is probably the sane choice to have here (which should be the default one). 

Screenshot 2024-01-28 at 12 29 35

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,6 +1,14 @@
 name: linkding CI
 
-on: [push]
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "*"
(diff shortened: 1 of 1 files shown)


sissbruecker:
Thanks, for now I've just configured:
```
on: [push, pull_request]
```
Which should be sufficient. Let me know if you think something else should be configured.

sissbruecker:
Welp, now it runs everything twice (push + pull request) when I open a PR. Need to spend more time on this.

PR #624, opened 2024-02-04
Insert 'Show more' if description exceeds two lines.

jonathan-s:
Similar to the title, the description can easily become truncated. So this patch separates tags and description into two different lines, and if the description becomes truncated there is an option to click "Show more" among the actions to see the full description.

### DIFF
--- a/bookmarks/frontend/behaviors/bookmark-page.js
+++ b/bookmarks/frontend/behaviors/bookmark-page.js
@@ -71,6 +71,28 @@ class BookmarkItem {
     if (titleSpan.offsetWidth > titleAnchor.offsetWidth) {
       titleAnchor.dataset.tooltip = titleSpan.textContent;
     }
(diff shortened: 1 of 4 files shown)


sissbruecker:
Uhm, so there's several things here. For one it changes the default UX so that the description is always on a separate line, which I'd like to avoid. Lots of people have been using linkding for years and are happy with how it works, myself included. So this has to be a setting.

The other thing is adding another button for this to the bookmark actions, which is already stuffed. It would also feel more natural if the button would show up at the end of the description somehow, ideally inline at the end of the text block. Not sure if that's possible though.

sissbruecker:
With https://github.com/sissbruecker/linkding/pull/663 and https://github.com/sissbruecker/linkding/pull/665 I'm inclined to close the PR. Let me know if you still think there is some value in this.

PR #630, opened 2024-02-16
Improve PWA capabilities

hugo-vrijswijk:
Improves the manifest.json to make the app installable on devices.

- Application is now "installable" and recognized as a PWA
- Shortcuts are added for common actions (if there's more that should be added, feel free to mention them)
- Add a `share_target` that allows sharing links to linkding when the PWA is installed (this works really well!)
- Updates and adds icons to improve installability on devices. Such as multiple sizes and "maskable" icons for dynamic icons on Android (that have a custom shape)
- Add application screenshot to the manifest.json so (on desktop) users will see a preview of the app when installing it


hugo-vrijswijk:
@sissbruecker could you take a look at this? :)

bookmarks/static/favicon.svg
sissbruecker, line 1: What's the benefit of replacing the existing favicon? Why have two versions of it (svg / ico)?

bookmarks/tests/test_metadata_view.py
```diff
@@ -26,8 +108,90 @@ def test_manifest_respects_context_path(self):
         response_body = response.json()
         expected_body = {
             "short_name": "linkding",
+            "name": "linkding",
```
sissbruecker, line 111: Nit: Instead of duplicating everything, maybe extract the common parts and reuse them in individual tests.

bookmarks/views/manifest.py
```diff
@@ -5,9 +5,91 @@
 def manifest(request):
     response = {
         "short_name": "linkding",
+        "name": "linkding",
+        "description": "Self-hosted bookmark service",
         "start_url": "bookmarks",
         "display": "standalone",
         "scope": "/" + settings.LD_CONTEXT_PATH,
+        "theme_color": "#5856e0",
+        "background_color": "#ffffff" if request.user_profile.theme == "light" else "#161822",
```
sissbruecker, line 14: This will also use the dark color when using the default setting "auto". Maybe switch the condition to check for dark specifically.

sissbruecker (commented):
Thanks for the effort, added a few comments.

bookmarks/static/favicon.svg
hugo-vrijswijk: ico's are fairly small. That means on higher resolution displays they don't scale as well when displayed larger. The svg is also a smaller size than the ico.

.dockerignore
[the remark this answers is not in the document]
hugo-vrijswijk: I've reverted the changes and submitted them separately in https://github.com/sissbruecker/linkding/pull/649

bookmarks/views/manifest.py
hugo-vrijswijk: Done, it's now white by default, only setting when the theme is dark specifically

bookmarks/tests/test_metadata_view.py
sissbruecker: Just noticed that a lot of URLs are different. Nevermind then, can be improved if more tests are added. As it is it's a pretty decent snapshot test.

dahamsta:
I'm still not getting this on Firefox, only the standard "Add to Home Screen". Are other people getting Install in the menu or Is it just a quirk of my install? I'm on Nightly.

hugo-vrijswijk:
> I'm still not getting this on Firefox, only the standard "Add to Home Screen". Are other people getting Install in the menu or Is it just a quirk of my install? I'm on Nightly.

I do get "Install" as an option on Firefox Android. Though I've installed it via Chrome because only Chrome supports the share target and shortcuts

leepeuker:
> > I'm still not getting this on Firefox, only the standard "Add to Home Screen". Are other people getting Install in the menu or Is it just a quirk of my install? I'm on Nightly.
> 
> I do get "Install" as an option on Firefox Android. Though I've installed it via Chrome because only Chrome supports the share target and shortcuts

I can confirm this, works on Android (14) with Firefox and Chrome

dahamsta:
Just occurred to me to test in standard Firefox, Install appears there but still not in Nightly, even with all extensions disabled. Would someone mind confirming please?

https://play.google.com/store/apps/details?id=org.mozilla.fenix

PR #633, opened 2024-02-17
Add k8s setup to community section

jzck:
Hi there, just wanted to share my way of hosting linkding. Thanks for the software!


sissbruecker:
Thanks! I've moved the link to the community section for now as it's not quite as hands-on or managed as the rest of the options. The definition of "Managed hosting options" is arguably a bit vague at the moment, maybe I'll improve that at some point.

PR #636, opened 2024-02-20
Normalize bookmark url metadata

jonathan-s:
I understand that you may be totally uninterested in this change. If that is the case, feel free to close this PR. 

This PR consists of three commits

1. First, adding the Link model, and adding the foreign key to bookmark. 
2. Second, migrate the data over to the Link model. 
3. Third, we remove the old fields in Bookmark change the code to only use data from the Link model.

If you have many users on the instance, it's a lot nicer to have normalized data as that data can then be re-used across those users, as it's not user specific.


sissbruecker:
Thanks for you effort, but I think I'd prefer not going ahead with this and keep the data model simpler. I can't see a lot of benefits to this, presumably there won't be a lot of overlap between user's bookmarks, so the reuse is marginal. On the downside this opens up a lot of potential for new bugs, and makes it harder to maintain in general.

PR #638, opened 2024-02-24
Added a new Linkding client to community section

JGeek00:
Hi. I'm the developer of [Linkdy](https://github.com/JGeek00/linkdy), a Linkding client created with Flutter. The application is still under development, and on a very early stage. The objective is to create a mobile and desktop client, with the same features that the web already has. I have added a new entry on the community section.


sissbruecker:
Thanks!

PR #644, opened 2024-03-06
Hardcode the default request timeout

ronau:
As described in [Options.md](https://github.com/sissbruecker/linkding/blob/master/docs/Options.md), the default value for `LD_REQUEST_TIMEOUT` is 60.
However, when you have something like this in your .env file

```
# Configures the request timeout (in seconds) in the uwsgi application server. Default: 60
LD_REQUEST_TIMEOUT=
```

then the variable's value will be unset (or null?).
In this case, uwsgi `http-timeout` seems to be [considered zero](https://github.com/sissbruecker/linkding/blob/38204c87cf6fdd3d871a115d9786db73b17d1be6/uwsgi.ini#L21), which will result in every request being dropped.

Using bash variable expansion rules, the default value 60 can be set as fallback.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -2,6 +2,7 @@
 # Bootstrap script that gets executed in new Docker containers
 
 LD_SERVER_PORT="${LD_SERVER_PORT:-9090}"
+export LD_REQUEST_TIMEOUT="${LD_REQUEST_TIMEOUT:-60}"
 
 # Create data folder if it does not exist
 mkdir -p data


sissbruecker:
Can you clarify how you ended up with a configuration like this? The setting is not in the .env sample file so you will not end up with this configuration if you start from there.

If it's unlikely that someone runs into this issue, I'd prefer not to do anything about it actually. Instead the setting should probably rather be removed, it doesn't really serve a purpose IMO. Originally the bookmark import was very slow, so that request could take longer than a minute, but now I can't think of any use-case where a request might take that long. Well, maybe if you're doing something crazy with the API.

ronau:
I tend to write my own .env files when incorporating new tools into my homelab collection. 😉
I usually include all the environment variables supported by the container, so that I can look them up easily right in the env file.
So yes, the problem is kind of my own fault. However, I think it makes sense to mention in the documentation that this value always must be an integer value.

If you think removing the setting entirely instead makes sense, then of course this is obsolete.

sissbruecker:
Thanks, for now I think I'll leave things as is. If this happens more often we can reconsider.

PR #649, opened 2024-03-16
build improvements

hugo-vrijswijk:
- Updates CI to run on Node 20 and updates Svelte to 4 (though there are no real differences)
- Use `npm ci` instead of `install` for CI and use npm cache for faster installs
- Update rollup versions
- Fix svelte accessibility warnings
- Use .mjs for rollup config (required since latest version of rollup)
- Remove unused/deprecated rollup plugins
- Update license field to correct license in package.json


hugo-vrijswijk:
Base is #630, should be merged first

bookmarks/frontend/components/SearchAutoComplete.svelte
```diff
@@ -209,9 +209,9 @@
     {/if}
     {#each suggestions.tags as suggestion}
       <li class="menu-item" class:selected={selectedIndex === suggestion.index}>
-        <a href="#" on:mousedown|preventDefault={() => completeSuggestion(suggestion)}>
+        <button on:mousedown|preventDefault={() => completeSuggestion(suggestion)}>
```
sissbruecker, line 212: Fixing this is unfortunately not as trivial:
Bildschirmfoto 2024-03-16 um 15 37 41

The CSS library used here explicitly targets `.menu-item a`,  using any other element as child will not style it properly.

sissbruecker (commented):
Thanks for doing these chores. The changes in the Svelte components are not as simple unfortunately. I would be fine with leaving them out for now.

bookmarks/frontend/components/SearchAutoComplete.svelte
hugo-vrijswijk: Oops, my bad. I should've tested this more. I've reverted it now

PR #672, opened 2024-04-01
Archive snapshots of websites locally

sissbruecker:
Allows to create HTML snapshots of websites and store them locally in the linkding data folder. This feature is only included in a new variant of the Docker image (`latest-plus`). I did not include it in the default variant as it consumes more resources: it requires a Chromium installation in the Docker image, more runtime memory to run Chromium, and more disk space to actually store snapshots. 

Snapshots are created using [single-file-cli](https://github.com/gildas-lormeau/single-file-cli). For now, snapshots can only be created on the server, which means websites with paywalls, or requiring login will not work. Future improvements could be to integrate this with the browser extension or allow uploading snapshots manually.

Snapshots are created automatically by default when using the alternative Docker image variant, a setting allows turning that off. Alternatively, snapshots can be created manually from the bookmark details view. Snapshots can be viewed from the bookmark details view as well.

Bildschirmfoto 2024-04-01 um 15 01 14


dahamsta:
Can you just switch to this image in docker-compose.yml, or does it need to be a fresh install?

sissbruecker:
You should be able to switch between images without issues. The default image just has the snapshot feature disabled and lacks the dependencies to run it.

PR #684, opened 2024-04-07
Add option to disable SSL verification for OIDC

akaSyntaax:
Adds the setting OIDC_VERIFY_SSL which allows to accept ssl certificates from untrusted (e.g. self-signed or created using a custom certificate authority) sources.

Without this setting, every OIDC login attempt fails with:
`requests.exceptions.SSLError: HTTPSConnectionPool(host='auth.example.com', port=443): Max retries exceeded with url: /api/oidc/token (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1006)')))`


sissbruecker:
Thanks!

PR #701, opened 2024-04-15
Add OCI source annotation to link back to source repo

Ramblurr:
This commit adds the `org.opencontainers.image.source` label to the built container images.

This label is helpful for tools to be able to link back from the container image to the source repo.

For example, for those that use Renovate to help auto update dependencies, this will result in the latest releases release notes/changelog being included in the PR which is very handy!

### DIFF
--- a/docker/alpine.Dockerfile
+++ b/docker/alpine.Dockerfile
@@ -69,6 +69,7 @@ RUN wget https://www.sqlite.org/${SQLITE_RELEASE_YEAR}/sqlite-amalgamation-${SQL
 
 
 FROM python:3.11.8-alpine3.19 AS linkding
+LABEL org.opencontainers.image.source="https://github.com/sissbruecker/linkding"
 # install runtime dependencies
 RUN apk update && apk add bash curl icu libpq mailcap libssl3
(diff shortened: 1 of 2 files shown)


docker/alpine.Dockerfile
```diff
@@ -127,3 +127,5 @@ COPY --from=ublock-build /etc/linkding/uBlock0.chromium uBlock0.chromium/
 RUN mkdir -p chromium-profile && chown -R www-data:www-data chromium-profile
 # enable snapshot support
 ENV LD_ENABLE_SNAPSHOTS=True
+
+LABEL org.opencontainers.image.source="https://github.com/sissbruecker/linkding"
```
sissbruecker, line 131: Thanks! Since the Dockerfiles contain multiple build stages that are used to generate an image (`linkding` and `linkding-plus`), I assume that the `LABEL` either needs to be set for both, or on the `linkding` build stage, which is the basis for `linkding-plus`.

Ramblurr: Oh you're right. I just assumed the last stage was the published one. I'll edit my commit.

PR #705, opened 2024-04-17
Updated options file with singlefile options and page restructure

ab623:
I noticed the Options doc did not contain a number of single-file options, and some option defaults were incorrect. 

I have updated them as best as I can, and also added a bit more structure into the document, given the number of options.

### DIFF
--- a/docs/Options.md
+++ b/docs/Options.md
@@ -25,21 +25,78 @@ All options need to be defined as environment variables in the environment that 
 
 ## List of options
 
-### `LD_SUPERUSER_NAME`
+## Database Options
 
-Values: `String` | Default = None
+### `LD_DB_ENGINE`
 
-When set, creates an initial superuser with the specified username when starting the container.
-Does nothing if the user already exists.
+Values: `postgres` or `sqlite` | Default = `sqlite`
 
-See [`LD_SUPERUSER_PASSWORD`](#ld_superuser_password) on how to configure the respective password.
+Database engine used by linkding to store data.
+Currently, linkding supports SQLite and PostgreSQL.
+By default, linkding uses SQLite, for which you don't need to configure anything.
+All the other database variables below are only required for configured PostgresSQL.
 
-### `LD_SUPERUSER_PASSWORD`
+### `LD_DB_DATABASE`
 
-Values: `String` | Default = None
+Values: `String` | Default =  `linkding`
 
-The password for the initial superuser.
-When left undefined, the superuser will be created without a usable password, which means the user can not authenticate using credentials / through the login form, and can only be authenticated using proxy authentication (see [`LD_ENABLE_AUTH_PROXY`](#ld_enable_auth_proxy)).
+The name of the database. 
+
+### `LD_DB_USER`
+
+Values: `String` | Default =  `linkding`
+
+The name of the user to connect to the database server.
+
+### `LD_DB_PASSWORD`
+
+Values: `String` | Default =  None
+
+The password of the user to connect to the database server.
+The password must be configured when using a database other than SQLite, there is no default value.
+
+### `LD_DB_HOST`
+
+Values: `String` | Default =  `localhost`
+
+The hostname or IP of the database server.
+
+### `LD_DB_PORT`
+
+Values: `Integer` | Default =  None
+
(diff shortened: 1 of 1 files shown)


docs/Options.md
```diff
@@ -25,21 +25,78 @@ All options need to be defined as environment variables in the environment that
 
 ## List of options
```
sissbruecker, line 26: Let's remove this one, seems superfluous with the new categories.

docs/Options.md
```diff
@@ -25,21 +25,78 @@ All options need to be defined as environment variables in the environment that
...
+The port of the database server.
+Should use the default port if left empty, for example `5432` for PostgresSQL.
+
+### `LD_DB_OPTIONS`
+
+Values: `String` | Default = `{}`
+
+A json string with additional options for the database. Passed directly to OPTIONS.
+
+### `LD_FAVICON_PROVIDER`
```
sissbruecker, line 77: Let's move this out of database options. I'd suggest to have a "Other options" or some other catch all category.

docs/Options.md
```diff
@@ -25,21 +25,78 @@ All options need to be defined as environment variables in the environment that
...
+
+Alternative favicon providers:
+- DuckDuckGo: `https://icons.duckduckgo.com/ip3/{domain}.ico`
 
-Values: `String` | Default = None
 
-The password for the initial superuser.
-When left undefined, the superuser will be created without a usable password, which means the user can not authenticate using credentials / through the login form, and can only be authenticated using proxy authentication (see [`LD_ENABLE_AUTH_PROXY`](#ld_enable_auth_proxy)).
+
+## Background Tasks
```
sissbruecker, line 99: This also contains a number of things not related to background tasks. Maybe repurpose this as the "Other options" category. Would also make sense to move this to the bottom of the doc then.

docs/Options.md
```diff
@@ -180,86 +239,69 @@ Values: `true` or `false` | Default =  `false`
...
-
-### `LD_DB_DATABASE`
 
-Values: `String` | Default =  `linkding`
+## Single File Options
 
-The name of the database. 
 
-### `LD_DB_USER`
+### `LD_ENABLE_SNAPSHOTS`
```
sissbruecker, line 246: This should not be exposed to users, it's only an internal marker for the app to know whether the dependencies for creating snapshots are installed and can be used. Changing this in the base image would result in broken features. Let's remove this.

docs/Options.md
```diff
@@ -180,86 +239,69 @@ Values: `true` or `false` | Default =  `false`
...
+```
+--browser-arg="--headless=new"
+--browser-arg="--user-data-dir=./chromium-profile"
+--browser-arg="--no-sandbox"
+--browser-arg="--load-extension=uBlock0.chromium"
+```
 
-Alternative favicon providers:
-- DuckDuckGo: `https://icons.duckduckgo.com/ip3/{domain}.ico`
+## Superuser Options
```
sissbruecker, line 291: These might be the most used options, which is why I had it at the top. Let's move it back there.

docs/Options.md
```diff
@@ -180,86 +239,69 @@ Values: `true` or `false` | Default =  `false`
...
+When creating HTML archive snapshots, pass additional options to the `single-file` application that is used to create snapshots.
+See `single-file --help` for complete list of arguments, or browse source: https://github.com/gildas-lormeau/single-file-cli/blob/master/options.js
 
-Values: `String` | Default =  `localhost`
+Example: `LD_SINGLEFILE_OPTIONS=--user-agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:124.0) Gecko/20100101 Firefox/124.0"`
 
-The hostname or IP of the database server.
 
-### `LD_DB_PORT`
+### `LD_SINGLEFILE_PATH`
```
sissbruecker, line 263: This probably shouldn't be changed either. If there's no use case for changing this it seems better to remove it to reduce confusion for readers what they should do with it.

sissbruecker (commented):
Thanks, clustering the options a bit is a good idea. Added a few comments.

sissbruecker:
Closing as stale

PR #710, opened 2024-04-17
Display descriptions with linebreaks correctly

### LINKED ISSUE #709: description field to support basic text formatting
I recently got a linkding instance set up for myself and an issue I've run into is that the description field formats all text as a single unbroken paragraph. I imported a large number of bookmarks where having hard returns in my descriptions is important in being able to easily read the information I put there. Having this field in the bookmarks list respect the returns that are in the input form would be very helpful.

ab623:
Fixes #709.

- On the details page, linebreaks will always be shown if the description has them
- On the main bookmarks index page, linebreaks will only be shown if `Bookmark description` is set to `Separate` (description will still be truncated if it contains more lines then the `Bookmark description max lines` setting

### DIFF
--- a/bookmarks/templates/bookmarks/bookmark_list.html
+++ b/bookmarks/templates/bookmarks/bookmark_list.html
@@ -47,7 +47,7 @@
         {% else %}
           {% if bookmark_item.description %}
             <div class="description separate">
-              {{ bookmark_item.description }}
+              {{ bookmark_item.description|linebreaks }}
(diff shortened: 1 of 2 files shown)


sissbruecker:
Thanks, but as explained in https://github.com/sissbruecker/linkding/issues/709#issuecomment-2067633304, it's not going to be as simple, and I'd hold off for adding a feature for this for now.

PR #713, opened 2024-04-20
Allow uploading custom files for bookmarks

sissbruecker:
Allows to upload custom files for each bookmark. That allows storing any arbitrary data (PDFs, images, text files, ...) that is related to that bookmark. Could also be used to store custom snapshots of a website in case the automated HTML snapshot does not work properly.

Bildschirmfoto 2024-04-20 um 11 52 33

### DIFF
--- a/bookmarks/frontend/behaviors/form.js
+++ b/bookmarks/frontend/behaviors/form.js
@@ -40,5 +40,25 @@ class AutoSubmitBehavior extends Behavior {
   }
 }
 
+class UploadButton extends Behavior {
+  constructor(element) {
+    super(element);
+
+    const fileInput = element.nextElementSibling;
+
+    element.addEventListener("click", () => {
+      fileInput.click();
+    });
+
+    fileInput.addEventListener("change", () => {
+      const form = fileInput.closest("form");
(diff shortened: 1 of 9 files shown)


ados8:
I'm unable to figure out how you can upload manual snapshots and have it display them not download.
I've even copied a snapshot from another website and uploaded that but it's a file only.
My assumption is the database needs to know it.
Could we get a feature to trigger not only a snapshot but provide the URL.
Use case, website 1 is working and there's a snapshot if not but the link on that page is dead.
Could use the WayBack Machine to get the dead link page and then make that a nested snapshot to the website in LINKDING.
Currently you have to download a PDF or manual HTML zip.

shazron:
I had trouble with a stanford.edu website and keep getting HTTP 403. I couldn't figure out what arguments to use for the singlefile-cli, in the end I used the browser extension: https://github.com/gildas-lormeau/SingleFile?tab=readme-ov-file#install which is available for all major browsers including Safari. 

@ados8 Try that - it downloads everything into one .html file, and you can upload that. Then it should display that HTML you uploaded when you click on the `View` link.

PR #724, opened 2024-05-09
Load bookmark thumbnails after import

vslinko:


vslinko:
it's ready to merge

bookmarks/services/preview_image_loader.py
```diff
@@ -31,15 +31,61 @@ def load_preview_image(url: str) -> str | None:
...
         return None
 
-    logger.debug(f"Loading preview image: {metadata.preview_image}")
-    with requests.get(metadata.preview_image, stream=True) as response:
-        content_type = response.headers["Content-Type"]
+    image_url = metadata.preview_image
+    content_length = 0
+
+    logger.debug(f"Loading preview image: {image_url}")
+    with requests.head(image_url, allow_redirects=True) as response:
```
sissbruecker, line 38: Is there any benefit to doing a separate HEAD request vs. doing the validation below in the GET request itself? Theoretically both requests could return different content lengths / types and status codes, so if those should be validated then it would make more sense to do so for the actual request that loads the image.

bookmarks/tests/test_website_loader.py
sissbruecker, line 1: Change here seems unnecessary.

bookmarks/services/preview_image_loader.py
sissbruecker, line 1: Would be great if you could add a few tests for the new validation conditions.

vslinko:
@sissbruecker fixed all your comments, could you please check again

sissbruecker:
Thanks again!

sissbruecker:
Also added this: https://github.com/sissbruecker/linkding/pull/725

PR #734, opened 2024-05-14
Thumbnails lazy loading

vslinko:
Prefer IntersectionObserver over loading=lazy because Safari still loads images when loading=lazy


sissbruecker:
Seems to work fine for me with Safari 16.6. Which version are you using?


https://github.com/sissbruecker/linkding/assets/357820/b22e467b-c63a-46aa-b7b0-32a0d1b1e0ab

vslinko:
I've checked again - looks like it works fine in current versions of Safari.

sissbruecker:
Alright, thanks.

PR #736, opened 2024-05-14
Automatically add tags to bookmarks based on URL pattern

vslinko:
Allows to automatically add tags to bookmarks by defining rules that map URLs to one or more tags. For example:
```
youtube.com video
reddit.com/r/Music music reddit
```


dsl.py
sissbruecker, line 1: I think I'd be fine with avoiding to build a custom UI for this for now, but it would be nice if the syntax would make it easy to add one later on. It seems that would be more challenging with a complex syntax that involves keywords and branches. It also seems more complex than it needs to be, thus making it harder to use. 

What about a simpler solution where each text line is a separate mapping? Each line would start with a domain, followed by a separator. After the separator you can specify any number of tags. Or maybe use JSON, which is more effort to write, but also more familiar, and even easier to modify through a custom UI later on.

vslinko: Maybe concider YAML?

```yaml
auto_tagging:
  - starts_with: https://www.youtube.com
    tags: [video]
```

sissbruecker: Maybe, though even as a developer I'm regularly challenged by YAML syntax and proper indentation. Would also require adding a dependency. Then there's things like having to declare a top-level section and correct object keys, which make it more error-prone.

I still think the simplest approach is per line:
```
www.youtube.com tag1 tag2 tag3
store.steampowered.com tag4 tag5
```

Alternatively using TOML, which is built in with Python 3.11, you could do:
```
"www.youtube.com" = ["tag1", "tag2", "tag3"]
```
Though that doesn't seem to provide any benefit to the above.

Regarding matching, since you used `starts_with`, do you think the matching needs to be this specific? Would just using the domain not be good enough?

vslinko: > Regarding matching, since you used starts_with, do you think the matching needs to be this specific? Would just using the domain not be good enough?

I have cases where domains are not enough. For example, I want to add a `music` tag for links starting with https://www.reddit.com/r/Music/.

By default it can be `contains`, but it will capture strings like `some-website.com/?q=youtube.com`.

I have an idea how to make the algorithm simple and flexible. Please see `dsl2.py`.

sissbruecker: Looks good 👍

Will push some changes and rebase shortly.

sissbruecker:
Works great, thanks again!

sissbruecker:
Created this as a follow up: https://github.com/sissbruecker/linkding/pull/737

I was thinking that it would be nice if it would just pre-fill the tags input with the auto tags. But then the user would also be able to remove automatically added tags and would expect them to not get added again when saving. That would get somewhat complex when trying to make this work through the extension and the REST API. Maybe I'll give that a shot later.

PR #739, opened 2024-05-19
Sort by last accessed

vslinko:
@sissbruecker Hi! What do you think of such a feature in the mainstream?

### DIFF
--- a/bookmarks/api/routes.py
+++ b/bookmarks/api/routes.py
@@ -12,6 +12,7 @@ from bookmarks.api.serializers import (
 )
 from bookmarks.models import Bookmark, BookmarkSearch, Tag, User
 from bookmarks.services.bookmarks import (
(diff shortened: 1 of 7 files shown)


vslinko:
@sissbruecker waiting for your feedback

garymoon:
@vslinko They might not be seeing this because it's a draft, I'd suggest marking it ready.

sissbruecker:
Thanks for the contribution, but I'm not sure if I want to proceed with this as I don't see a lot of utility. I'd suggest to create an issue first and get some feedback if others are interested in this and what their use cases would be. I'll still keep this open for now.

Regarding the implementation, I'm not sure if this approach works reliably when opening bookmarks in the same page, though that would require some testing. The alternative might be to intercept link clicks, make the API request first, and then navigate programmatically.

Issue #764, opened 2024-06-20
Date and time incorrect


ados8:
I've added the TZ variable to the Unraid docker template but it continues to show the time and therefore date far behind.
At first I thought it was the snapshot system as it might have it's own timezone but even the bookmark adding is out.
Is there a way to set the timezone?

watsonbox:
I tried the same thing with no luck, but [this solution](https://github.com/sissbruecker/linkding/issues/179#issuecomment-986292357) worked by updating `docker-compose.yaml`:

```yaml
volumes:
  - "${LD_HOST_DATA_DIR:-./data}:/etc/linkding/data"
  - "./custom.py:/etc/linkding/siteroot/settings/custom.py:ro"
```

I note that there is another proposal to solve this in the app itself: https://github.com/sissbruecker/linkding/issues/722.

After playing with it for a bit, I decided to submit a small fix to get the `TZ` var working  https://github.com/sissbruecker/linkding/pull/765.

Issue #179, opened 2021-12-04
Timezone does not change when changing a variable "TZ" in docker


molejado:
Really like the image, but I have the problem of changing the time zone. When i change the variable tz = Europe/Moscow, server ignores timezone plus 3 and all bookmarks saves with a timezone minus 3 hours. Please tell how to fix this bug. 
Sorry if I presented the problem incorrectly, thank you for linkding.

![123](https://user-images.githubusercontent.com/95546757/144723256-a7541ae1-ace0-42b3-b6a5-2ebbe8c13322.png)

sissbruecker:
Not sure where you got the `tz` variable from, but it's not something that is supported. linkding uses Django, which has a `TIME_ZONE` variable that can be set in the settings. By default, that variable is set to UTC: https://github.com/sissbruecker/linkding/blob/master/siteroot/settings/base.py#L118

You can change that by mounting a [custom settings file](https://github.com/sissbruecker/linkding/blob/master/siteroot/settings/custom.py) into the container at `/etc/linkding/siteroot/settings/custom.py`, and then setting the `TIME_ZONE` variable to something else:
```python
TIME_ZONE = 'Europe/Moscow'
```

Setting the timezone per user is not supported ATM.

molejado:
Thank you, now the time zone is working correctly, I am very grateful to your project, I have a cup of coffee for you)

FunctionDJ:
It would be awesome if setting the time zone was supported via Docker enrivonment (e.g. docker compose)

wxyShine:
> 如果通过 Docker 环境支持设置时区（例如 docker compose），那就太棒了

Do you support setting time zone now?

PR #765, opened 2024-06-20
Allow use of standard docker `TZ` env var

watsonbox:
I noticed that the time zone wasn't set up right for me, and eventually came across [this](https://github.com/sissbruecker/linkding/issues/179#issuecomment-986292357) which got me up and running.

However, both that issue and [this more recent one](https://github.com/sissbruecker/linkding/issues/764) were opened by people trying to use the standard docker `TZ` var to set the app time zone. That seems like a reasonable thing to do.

If you agree, what do you think of this small fix?

> [!NOTE]
> An alternate approach might be to use the system time zone instead of UTC as the default. That would give the benefit of working for both bare metal installs as well as docker containers configured with `TZ`.

Closes https://github.com/sissbruecker/linkding/issues/764

### DIFF
--- a/siteroot/settings/base.py
+++ b/siteroot/settings/base.py
@@ -114,7 +114,7 @@ LOGOUT_REDIRECT_URL = "/" + LD_CONTEXT_PATH + "login"
 
 LANGUAGE_CODE = "en-us"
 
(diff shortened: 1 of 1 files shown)


sissbruecker:
Thanks, seems like a good start

Issue #659, opened 2024-03-19
Tag bundles


zerodogg:
There's one feature of pinboard that I found rather useful, and that's "tag bundles". It lets you browse your bookmarks using these bundles, which makes it possible to narrow down the view without having to go down into a single tag.

As an example, I had a "Healthcare" bundle (I studied, and now work in healthcare), that bundles together tags like "anotamy", "health", "psychiartry" and so on, roughly 40-50 tags. When I clicked on the bundle, it narrowed the view down to any bookmark containing any one of these tags. I don't need to be seeing all my programming bookmarks when I'm after healthcare info. I had like 10 different bundles, including things like "Development", "Games", "Science".

This can be solved by adding yet-another tag, of course. But this felt like a cleaner abstraction than doing that.

This is somewhat related to #205.

Thanks for a great bookmark manager!

sissbruecker:
Sounds very much like https://github.com/sissbruecker/linkding/issues/285#issuecomment-1213905426

That doesn't mention filtering by the category / bundle / tag group or whatever it would be called, but for me that is implicit. Maybe give that comment an upvote if you prefer that style of categorization. I guess we can leave this issue open to see if this gets some more reactions than the comment in the other thread.

hojerst:
i personally would like to see "hierarchical tags" which is probably very similar to your request. (I'm not sure if it fits your needs though). Some other tools Have special meaning for the `/` character in tags and i really like this approach, as it is quite simple from a user interface perspective:

The idea is to tag your stuff with something like `category/tag` (maybe even more nesting). Every search should allow to find this item by `category` or by `category/tag`

example: adding tag `world/europe` lets you search by `world` and `world/europe`.

This could probably be implement with a reasonable amount of effort in the ui and backend (i'm just guessing here, as i don't know the internals of linkding at all ;)). @sissbruecker whats your opinion about that? If you like the approach, i might try to create PR for that.

zerodogg:
@sissbruecker thank you very much for implementing this! 😊

PR #780, opened 2024-07-13
add hierarchical tag search

hojerst:
My shot at implementing a tag hierarchy at the search level for #659 

Tags with `/` characters can be searched by any of its parents tags (i.e. a tag `a/b/c` is searchable via `a`, `a/b` or `a/b/c`.

Added an option to enable/disable this functionality in the user profile (default is off - aka old behaviour).


anb-apg:
Now this would be a great feature. I didn't have a look at the code, but one would have to search for `#b`, not `b`, right?

hojerst:
yes, normally you would search for tags via "#" prefix as normal.

it actually depends on your "tag search" setting as before. if it is set to "lax" it means you can search for tags without "#" as well. so no changes at the point.

sissbruecker:
Thanks for the contribution. There definitely should be a better way to categorize bookmarks, but I'm not sure if this is it. The good thing about this implementation is that it's opt-in and a rather isolated change. Personally, I'd want something different for tag bundles though, in that I can define a bundle like `programming` which:
- Is selectable through the UI, separately from the cluttered tag cloud
- Allows me to define a number of tags like `css`, `java`, ... that are automatically included in the search results when searching for `programming`

I'll keep this open as an option, but I still have to make up my mind if categorization should be solved with folders, tag bundles, tag hierarchies, or by supporting multiple options. Or possibly a combination of those, for example folders could be combined somewhat with the tag bundle approach I mentioned above.

anb-apg:
> I'll keep this open as an option, but I still have to make up my mind if categorization should be solved with folders, tag bundles, tag hierarchies, or by supporting multiple options. Or possibly a combination of those, for example folders could be combined somewhat with the tag bundle approach I mentioned above.

(Folders are the root of all evil - my personal opinion) 

Could a bookmark then reside in multiple folders?

hojerst:
at least the main pull request here is nothing more then a partial tag search. this means you can organize your tags in a folder like hierarchy and then search for the path or the full tag name. this means you can have multiple hierarchical (or notmal) tags per link.

sissbruecker:
> (Folders are the root of all evil - my personal opinion)

I would agree with that. I hated it when every time I created a bookmark I had to decide in which folder to put it. Or digging through the folder hierarchy to find something. I much prefer just adding some tags and then relying on search. But people are used to them and it's one of the most requested features at this point. Apart from that the current tag cloud quickly becomes useless / too noisy if you want to navigate bookmarks through the UI rather than through search. So some form of a configurable curated list of categories (folders, tag bundles, whatever) as a quick filter makes sense.

> Could a bookmark then reside in multiple folders?

With tag bundles very likely, as tags don't have a hierarchy and you can assign multiple of them to a bookmark. If there would be a feature for manually assigning bookmarks to folders then I'd probably not go that far in a first version.

ccxuy:
> (Folders are the root of all evil - my personal opinion)
> 
> Could a bookmark then reside in multiple folders?

Kind of true, maybe put tags in folder, a tag belongs to some other tags (category), act like tags, but a bookmark still could be found via it's parent tag. Also when tag glows, the UI would become less messy

Bonjour123:
> Thanks for the contribution. There definitely should be a better way to categorize bookmarks, but I'm not sure if this is it. The good thing about this implementation is that it's opt-in and a rather isolated change. Personally, I'd want something different for tag bundles though, in that I can define a bundle like `programming` which:
> 
> * Is selectable through the UI, separately from the cluttered tag cloud
> * Allows me to define a number of tags like `css`, `java`, ... that are automatically included in the search results when searching for `programming`
> 
> I'll keep this open as an option, but I still have to make up my mind if categorization should be solved with folders, tag bundles, tag hierarchies, or by supporting multiple options. Or possibly a combination of those, for example folders could be combined somewhat with the tag bundle approach I mentioned above.

Yep, tag-bundle + tag synonyms would be just great !

sissbruecker:
Thanks again for the contribution, however with the introduction of bundles this I feel this is obsolete.

Issue #776, opened 2024-07-06
Allow configuring `/shared` as default page for unauthenticated users


thetiny99:
(The following content is translated by me using translation engine) 

Hello, I saw LINKDING by chance. No matter from its functions, interface or update stability, I feel like I have found a treasure. Thank you to the developers for their hard work!

The only regret is that LINKDING does not seem to support viewing content without logging in. In other words, you can view bookmarks as a "guest" without adding an account.

This demand scenario still exists:
1) Use LINKDING to build a public website and share bookmarks with a broad audience;
2) Use LINKDING to share public links in organizations, communities, schools, etc.

——LINKDING’s design in terms of privacy and authentication is impeccable. I hope developers can consider the anonymous browsing function. Thank you!

jburnham:
Hi @thetiny99,

I believe this feature already exists in some form and I am providing the steps on how to enable this feature below.

In Settings -> General you can enable public bookmark sharing by checking these two boxes and saving.
- Enable bookmark sharing
- Enable public bookmark sharing

![image](https://github.com/sissbruecker/linkding/assets/254022/b49b2897-f108-4d51-983e-2ae1395b9428)

When you edit a bookmark, you can then enable Share.
![image](https://github.com/sissbruecker/linkding/assets/254022/48e3b2b9-2c19-4f5f-b543-64870544d703)

The https://linkding.example.com/bookmarks/shared page will then show all shared bookmarks without needing a login.
![image](https://github.com/sissbruecker/linkding/assets/254022/782bda4e-b4bf-49b6-bdf5-e4e003fb1114)

This shared bookmark page is linked on the homepage of the app before logging in.
![image](https://github.com/sissbruecker/linkding/assets/254022/d38c4907-dc9f-4237-bc8a-2ca051f584c3)

There are also settings to automatically make imported bookmarks shared which you can see below.
![image](https://github.com/sissbruecker/linkding/assets/254022/c368cb5e-2f8c-483b-953c-9494fb97c46b)

Best of luck and I hope this feature as implemented is acceptable for you.

thetiny99:
[...]

PR #808, opened 2024-08-31
Allow configuring landing page for unauthenticated users

sissbruecker:
Adds a global settings form, that allows to choose which page unauthenticated guest users are redirected to when accessing the root URL. Options are the login page (default), or the shared bookmarks page.

Closes https://github.com/sissbruecker/linkding/issues/776


D1reW0lf:
Hi. 
I noticed this is not on 1.36.0.
Any idea when this will be available?

Thank you.

sissbruecker:
This has been released with [1.32](https://github.com/sissbruecker/linkding/releases/tag/v1.32.0). Note that your user needs admin permissions to see the global settings.

Issue #714, opened 2024-04-20
ability to configure the default profile used for anon users


soph-sol:
I'd love to have the ability to configure the default profile used for anonymous users, so that when people visit my shared bookmarks page, it's set up to display information in a way that's helpful for the kinds of things I bookmark.

Specifically I'd like to be able to set how many lines of the description are visible, as well as add custom CSS the way I can for my own user profile.

bloodripelives:
This would be great!

nuriarita:
I'd love that!

Issue #533, opened 2023-09-11
Shared bookmarks page displays notes by default


bfansheng:
Shared bookmarks page is a great feature. I use the shared bookmarks page to share my favorite articles with others and my thoughts on them, so I hope my bookmarks notes can be seen directly by others. currently, the notes on the shared bookmarks page are not displayed by default, which may make it difficult for others to notice my notes. I think it might be a good idea to have the notes displayed by default on the sharing page.

sissbruecker:
I think a setting would make sense here, and can think of at least 3 options:
- Show notes collapsed (current default)
- Show notes expanded / permanently
- Do not show notes at all in case you consider your notes private

keybits:
I'd love the option for Notes to be automatically expanded on the Shared Bookmarks page. This would allow me to use Linkding as a 'blogmarks' site like Simon Willison does: https://simonwillison.net/search/?type=blogmark

PR #809, opened 2024-08-31
Allow configuring guest user profile

sissbruecker:
Allows configuring a guest user profile in the global settings. The profile can be used to customize how the publicly shared bookmark list is displayed for guest users.

Closes https://github.com/sissbruecker/linkding/issues/714
Closes https://github.com/sissbruecker/linkding/issues/533


bloodripelives:
Just updated and this is great, thank you so much!

PR #821, opened 2024-09-11
Add API endpoint (`api/bookmarks/export`) to allow exporting the bookmarks in netscape html format

obrhubr:
I am not familiar at all with Django development, but I wanted this feature for personal use. Because your project doesn't really have any guidelines for contributing, I don't know what you expect from a pull request (or if you even accept any).

I would be willing to add to my changes to make my branch ready for merging.

### DIFF
--- a/bookmarks/api/routes.py
+++ b/bookmarks/api/routes.py
@@ -6,6 +6,9 @@ from rest_framework.permissions import AllowAny
 from rest_framework.response import Response
 from rest_framework.routers import DefaultRouter
 
+from django.http import HttpResponse
+from django.db.models import prefetch_related_objects
+
 from bookmarks import queries
 from bookmarks.api.serializers import (
     BookmarkSerializer,
@@ -20,6 +23,7 @@ from bookmarks.services.bookmarks import (
     website_loader,
 )
 from bookmarks.services.website_loader import WebsiteMetadata
+from bookmarks.services import exporter
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


radiant-tangent:
There is an export feature in settings that uses the `/settings/export` endpoint. How does this improve upon that functionality?

obrhubr:
The export endpoint under `/settings/export` is restricted to session based authentication. It's not a REST endpoint meant to be accesses programmatically by a script.
Session based auth is only possible through the CSRF protected login enpoint for linkding.
This PR improves on that by providing this functionality as a REST API.

sissbruecker:
I assume you want to use this for backups? I'm not sure if I want to encourage that by providing an option to automate this through the API. The export is quite limited in what it supports, and I'd rather encourage to use the actual backup mechanism described in the docs.

DonkeeeyKong:
> I assume you want to use this for backups? I'm not sure if I want to encourage that by providing an option to automate this through the API. The export is quite limited in what it supports, and I'd rather encourage to use the actual backup mechanism described in the docs.

I think, It would be nice to have this not in place of but in addition to a standard backup as an 'emergency backup' for the worst case scenario: For some reason linkding stops working and I can't get it back. 
In this scenario, if I had automated exports to a standardized format enabled, it would be easily possible to access my bookmarks from somewhere else or just import them quickly into e.g. my browser.
As the bookmarks are the most important item to be backed up, this would be a good fall-back option.

PR #826, opened 2024-09-15
Implement IPv6 capability

itz-Jana:
I've been wanting to start using linkding and installed it into my Kubernetes cluster, but couldn't access it.

My cluster runs only using IPv6 addresses and I noticed that the default configuration for uWSGI that is used in this project, defaults to binding only to an IPv4 socket and there is no config option to change this behaviour.

In the current year, with IPv4 addresses becoming more scarce and often an addon that needs to be paid extra, I think it's reasonable to change the default to **also** allow for IPv6 connections.

**This does not break any existing configurations**, as this creates a dual stack socket, which will respond to requests on both protocols (IPv4 addresses are displayed as IPv4-mapped IPv6 addresses and look like this: "::ffff:x.x.x.x".
Yet I have also provided a configuration option to change the behaviour, if a user were to desire it for whatever reason.

If you don't like changing the default, we could swap it back to 0.0.0.0 and only allow overwriting it, if desired, but as I have outlined I think changing the default is reasonable and doesn't harm existing setups.

I have tested these changes on my IPv6-only cluster, as well as on Docker Desktop on Windows with port mapping (which uses IPv4 addressing) and tried it on both the normal default and alpine image and it worked in all cases.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,9 +1,7 @@
 # Docker container name
 LD_CONTAINER_NAME=linkding
 # Port on the host system that the application should be published on
-LD_SERVER_HOST=[::]
-# Port on the host system that the application should be published on
-LD_SERVER_PORT=9090
+LD_HOST_PORT=9090
 # Directory on the host system that should be mounted as data dir into the Docker container
 LD_HOST_DATA_DIR=./data
 
--- a/.env.sample
+++ b/.env.sample
@@ -1,7 +1,9 @@
 # Docker container name
 LD_CONTAINER_NAME=linkding
 # Port on the host system that the application should be published on
-LD_HOST_PORT=9090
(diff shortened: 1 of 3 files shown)


sissbruecker:
Thanks, makes sense I guess. Seems to work fine for me. If people are having issues I might revert it, then we can try the alternative option-only approach you suggested.

sissbruecker:
Reverted the changes to the env sample. `LD_HOST_PORT` was actually correct, it's the port that is bound on the host system, whereas `LD_SERVER_PORT` is the one that uwsgi runs on. This is customizable for container runtimes that use a shared port range. Also removed the new variable from the sample, people should rarely need to customize this, so no need to confuse people with that option.

PR #836, opened 2024-09-18
Add go-linkding to community projects

piero-vic:
Hi. First of all, thanks for the work on linkding. I've developed [go-linkding](https://github.com/piero-vic/go-linkding), a go library to interact with the API. I'm in the process of adding some tests that run against the docker container to ensure the library is always compatible with the latest version of linkding. I'm already using it in a couple of projects. Hope the community can make good use of it.


sissbruecker:
Thanks!

PR #842, opened 2024-09-20
Show placeholder if there is no preview image

sissbruecker:


hotlcc:
I'm sorry, I don't think it looks very good like this

PR #848, opened 2024-09-23
Add bookmark count to header.

djl:


bookmarks/templates/bookmarks/index.html
```diff
@@ -9,7 +9,7 @@
     {# Bookmark list #}
     <section class="content-area col-2">
       <div class="content-area-header mb-0">
-        <h2>Bookmarks</h2>
+        <h2>Bookmarks ({{ bookmark_list.items | length }})</h2>
```
sissbruecker, line 12: `items` only contains the items for the current page, so at max. it would only show the number of items for the current page instead of for the overall search result.

bookmarks/templates/bookmarks/index.html
sissbruecker, line 1: While I can see this being useful, in general I'm hesitant to add new stuff to the bookmark list without making it optional. So I think this should be opt-in as well via the profile settings.

Also please look into writing unit tests for the new feature.

djl:
Hmm I didn't mean to submit this PR yet. Thanks, GitHub!

I'll mark this as draft for now.

sissbruecker:
I'll close this for now, feel free to open a new PR when you are ready.

PR #850, opened 2024-09-23
Fix header.svg text

vladh:
Currently, the logo in the README looks broken when Helvetica Neue is not installed:

linkding-logo-broken

This is because the text is not expanded to curves — at least, interestingly, “l” is expanded, but “inkding” is not.

The SVG in this PR contains expanded text based on the .afdesign file.

diff


sissbruecker:
Thanks! I hope I can remember to click that checkbox if I ever change that file again...

PR #860, opened 2024-09-27
Add LAST_MODIFIED attribute when exporting

ixzhao:


sissbruecker:
What would be the benefit of adding this?

sissbruecker:
Preserving information is not really a benefit in itself, unless it supports some other feature. The modified date isn't really used anywhere in the app, which is why I'm asking.

sissbruecker:
I saw that `LAST_MODIFIED` is something that's also supported by browsers, with that I'm fine with adding it.

One thing that's missing from this PR is tests, would be good if you could add those.

PR #871, opened 2024-10-05
Fix issue where description gets returned as null to extension

radiant-tangent:
Issue: Some websites, such as this [mastodon page](https://mastodon.social/@AJCxZ0@infosec.exchange/113205687391468478) have no description and so the `check` endpoint was returning `null` to the extension for the description field.  When the user goes to save, they get the message "Error saving bookmark: Validation error: {"description":["This field may not be null."]}".
This error does not occur when metadata is loaded from the browser instead of the server.

Solution: When a description is not found an empty string will be returned.

### DIFF
--- a/bookmarks/services/website_loader.py
+++ b/bookmarks/services/website_loader.py
@@ -60,6 +60,8 @@ def load_website_metadata(url: str):
                 else None
             )
 
+        description = description if description else ""
+        
         image_tag = soup.find("meta", attrs={"property": "og:image"})
         preview_image = image_tag["content"].strip() if image_tag else None
         if (


sissbruecker:
Thanks for catching that! It's a bit strange though that the API would return an empty string if no content was found at all. And it would be inconsistent with other fields like the preview image for example, where it makes a lot of sense to return `null` if nothing was found.

Maybe the extension should make sure that it's not sending `null` to the API instead? That seems like a good idea in general as that might also cover other cases in the future where the description might end up having an unexpected value.

radiant-tangent:
Is there a functional advantage to returning `null` from the API? Is there any situation that is making a decision based on it being `null`?  I considered changing the extension first but since all the other places that return description (browser metadata, existing bookmarks) return empty strings, this approach seemed the most consistent.  
I considered changing the preview image too but I didn't have a good way to test downstream effects and it wasn't causing any errors on my side with the way it is currently written.

sissbruecker:
> Is there a functional advantage to returning null from the API? Is there any situation that is making a decision based on it being null?

Not within linkding or the extension. But possibly in a 3rd party app or script that someone uses. Might be better to not change the behavior for that reason alone.

The issue here is the result of a change in the extension, so it seems better to deal with it there as well. Agree that it's inconsistent with what the extension does internally, but that seems like a lesser issue. Besides, the browser metadata logic could be changed to return null as well, and then the popup could deal with both cases similarly.

radiant-tangent:
Closing this since we've decided to change the extension instead. See https://github.com/sissbruecker/linkding-extension/pull/86

PR #880, opened 2024-10-11
Add serchding to community projects, sort the list by alphabetical order

ldwgchen:


sissbruecker:
Thanks, but can you please revert the sorting change - it doesn't make sense to sort uppercase and lowercase separately (if that is what you intended).

ldwgchen:
> Thanks, but can you please revert the sorting change - it doesn't make sense to sort uppercase and lowercase separately (if that is what you intended).

You're right - the commit should now reflect the reversion

PR #887, opened 2024-10-15
Add option to disable request logs

### LINKED ISSUE #785: Add `disable-logging` environment variable
# About

I have Linkding running in a k3s cluster, and I scrape all the pod logs into a central Loki cluster for searching etc. The logs are full of entries like this:

`localhost [pid: 4029|app: 0|req: 1/1] 127.0.0.1 () {48 vars in 906 bytes} [Wed Mar 23 18:35:38 2016] GET / => generated 94847 bytes in 1137 msecs (HTTP/1.1 200) 4 headers in 224 bytes (1 switches on core 0)`

The Kubernetes health-checker checks the pod health every minute or so, so my logs are filled with these entries. To reduce space and compute costs, I have disabled request logging in my `Dockerfile` by adding `disable-logging=True` to `uwsgi.py`.

You can enable logging for 40x and 50x by adding:

```
log-4xx = true
log-5xx = true
```

I think it'd be beneficial to add this to linkding

# Version
1.31.0 on armv8/linux

# Solution

Add:

```python
if-env = LD_DISABLE_LOGGING
disable-logging = true
log-4xx = true
log-5xx = true
endif =
```

to `uwsgi.py`.

dmarcoux:
Closes #785

### DIFF
--- a/docs/src/content/docs/options.md
+++ b/docs/src/content/docs/options.md
@@ -268,3 +268,9 @@ When creating HTML archive snapshots, pass additional options to the `single-fil
 See `single-file --help` for complete list of arguments, or browse source: https://github.com/gildas-lormeau/single-file-cli/blob/master/options.js
 
 Example: `LD_SINGLEFILE_OPTIONS=--user-agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:124.0) Gecko/20100101 Firefox/124.0"`
+
+### `LD_DISABLE_REQUEST_LOGS`
+
+Values: `true` or `false` | Default =  `false`
+
+Set uWSGI [disable-logging](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#disable-logging) parameter to disable request logs, except for requests with a client (4xx) or server (5xx) error response.
--- a/uwsgi.ini
+++ b/uwsgi.ini
@@ -29,3 +29,9 @@ endif =
 if-env = LD_LOG_X_FORWARDED_FOR
 log-x-forwarded-for = %(_)
 endif =
+
+if-env = LD_DISABLE_REQUEST_LOGS=true
+disable-logging = true
+log-4xx = true
+log-5xx = true
+endif =


dmarcoux:
I tried to test this locally by adding `env = LD_DISABLE_LOGGING=true` to `uwsgi.ini` and spinning up the server with `uwsgi --http :8000 --module siteroot.wsgi`. The logs were still being shown. I am not super familiar with uWSGI, maybe someone could guide me here?

sissbruecker:
It's an environment variable, so you either set it using `export LD_DISABLE_LOGGING=1` when testing locally, or adding `-e LD_DISABLE_LOGGING=1` when starting the Docker container.

The change seems to work. Naming isn't great, as it doesn't really disable logging:
- It only disables request logs
- It will still log failed requests
- It will still log stuff from linkding itself

Apart from that the option should be documented in the options docs.

dmarcoux:
My first attempt to test this was with `LD_DISABLE_LOGGING=1 python3 manage.py runserver`, but I could still see the request logs (for example from the `/health` endpoint).

Anyway, it's true that the naming isn't great. Maybe `LD_QUIETER_LOGGING`? Naming things is hard :smile:

sissbruecker:
> My first attempt to test this was with LD_DISABLE_LOGGING=1 python3 manage.py runserver

That command uses the Django dev server, and not uwsgi. I guess `LD_DISABLE_LOGGING=1 uwsgi --http :8000 --module siteroot.wsgi` should do it then.

Regarding naming, `LD_DISABLE_REQUEST_LOGS` would be more appropriate IMO, if not 100% correct. But could be fine if the docs mention that errors are still logged.

docs/src/content/docs/options.md
```diff
@@ -268,3 +268,9 @@ When creating HTML archive snapshots, pass additional options to the `single-fil
 See `single-file --help` for complete list of arguments, or browse source: https://github.com/gildas-lormeau/single-file-cli/blob/master/options.js
 
 Example: `LD_SINGLEFILE_OPTIONS=--user-agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:124.0) Gecko/20100101 Firefox/124.0"`
+
+### `LD_DISABLE_REQUEST_LOGS`
+
+Values: `true` or `false` | Default =  `false`
```
sissbruecker, line 274: Technically this seems to be active as long as the env variable is defined in some way, regardless of the value. So setting it to `false` will actually activate it. 

So maybe this should not list possible values, but rather state that whenever this env variable is defined with any value, then request logs will be disabled. Or investigate if the syntax in uwsgi.ini allows checking for specific values.

sissbruecker: This seems to do it:
```
if-env = LD_DISABLE_LOGGING=true
disable-logging = true
log-4xx = true
log-5xx = true
endif =
```

PR #909, opened 2024-11-12
Add active states to top-level nav menu items

1ockwood:
## Changes

This is a fairly simple change that adds active states to the top-level nav menu items on large viewports. As a new user, the first time I clicked around, I expected to see that sort of feedback to give a sense of where I am in the app. 

I kept it to the "desktop" version of the nav menu only, as the "mobile" menu is structured a bit differently (the top-level "Bookmarks" item being grouped at the same hierarchy as the other Bookmarks entries), and I don't necessarily feel it needs the same feedback in that context. Happy to experiment with that though if you feel otherwise. 

In the same vein, I also tweaked the styles for hover states for the nav menu links.

### Dark Mode

nav-menu-active-items-dark

https://github.com/user-attachments/assets/8359c87c-d4c5-4d63-a284-ed48cb78be42

### Light Mode

nav-menu-active-items-light

https://github.com/user-attachments/assets/ec1a5deb-9213-4b12-ab16-2a01ad776c82

### DIFF
--- a/bookmarks/styles/theme-dark.css
+++ b/bookmarks/styles/theme-dark.css
@@ -102,7 +102,7 @@
   --btn-error-text-color: var(--contrast-text-color);
 
   --btn-link-text-color: var(--link-color);
-  --btn-link-hover-text-color: var(--link-color);
+  --btn-link-hover-text-color: var(--secondary-link-color);
 }
 
 :root {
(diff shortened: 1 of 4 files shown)


tkafka:
I love this! Any chance to merge?

PR #910, opened 2024-11-12
Prefer html snapshots when internet archive integration is disabled

cything:
Right now the timestamp always links to the Internet Archive snapshot even when Internet Archive integration is disabled. I believe it generates a fallback URL that's close to the timestamp which does work well but I think it's confusing. When integration is disabled and there is a local HTML snapshot available, it makes sense to prefer that over the fallback wayback machine URL. Also, there's no other way to change this behavior AFAIK.

I was debating whether there should be a separate setting to change what clicking on the timestamp does, but I think people would expect that disabling Internet Archive integration shouldn't still link to it.


JamesMenetrey:
Hello @cything, many thanks for having addressed my issue!

@sissbruecker, do you think this PR can be merged? Thanks for your work!

cything:
> @sissbruecker, do you think this PR can be merged? Thanks for your work!

He hasn't been active in maintaining this repo so I wouldn't count on this getting merged soon. You can checkout this PR and deploy on your server if this feature is important to you.

sissbruecker:
Thanks, I took some cues from this PR and implemented an alternative with https://github.com/sissbruecker/linkding/pull/1021.

PR #914, opened 2024-11-17
Add a rust client library to community.md

zbrox:
Hey,

I added linkding-rs to the community projects list. It is a Rust client with some FFI bindings so it can be easily used in Kotlin/Swift if making a mobile app. 

I made myself a rudimentary Swift app with a share extension using it, so I can more easily save bookmarks to my linkding instance from mobile Safari or whichever other iOS browser.


sissbruecker:
Thanks!

PR #933, opened 2025-01-01
docs: add link to bookmarklet section

shuuji3:
Hi, this PR add a link in README to the bookmarklet page. This will help new users with the initial setup for Linkding.

Screenshot of README with red rectangle around "bookmarklet" text


sissbruecker:
This would only help if you are looking for how to set it up on mobile specifically. Otherwise it would be confusing users who want to set it up in a desktop browser. I think I'll keep this as is for now.

shuuji3:
That's fair. I think I should have thought about it more before sending PR. Thank you.

PR #949, opened 2025-01-15
Add 3 new community projects

sebw:
Archiving, reminding and healthchecking.


sissbruecker:
Thanks! Regarding linkding-archiver, the linkding plus image has support for archiving with Singlefile as well. For storing external links in the bookmark you could consider using the notes field which would allow you to render an actual link using markdown.

sebw:
I added it to the list but please feel free to remove that mention if you feel it is not relevant. My extension is lacking versioning compared to the built in feature.

I built archiver two years ago before you implemented the feature, and that was indeed using the notes section 😃

PR #972, opened 2025-01-30
add option to disable form login

### LINKED ISSUE #730: Disable Local Login After Enabling Authentik OIDC
Hi all, just wanted to see if this is an option somewhere, or if not, humbly request if it could be added. I've gotten Authentik working for SSO via OIDC for Linkding, and it seems to be working correctly (fingers crossed the mobile app and browser extension don't have errors).

However, I was hoping to bypass the login screen where it's merely an option to choose the OIDC login, and force it to the Authentik login page instead.

I know this is something requiring explicit toggling for other apps (Nextcloud and Bookstack for instance) - did not know if that was already the case here, and if so how.

kyuuk:
Fixes #730 

Add an option to disable login form, this hides the username and password fields and Login button, 
i also changed the `Login with OIDC` button css classes to make it a little better when form is disabled
which gives a for like this :
image

i'm not sure about the frontend part , it could probably be improved.

- Added test to validate that by default fiels are still present
- Added test to validate that fields are not present when option is true

### DIFF
--- a/bookmarks/templates/registration/login.html
+++ b/bookmarks/templates/registration/login.html
@@ -15,6 +15,7 @@
           <p class="form-input-hint">Your username and password didn't match. Please try again.</p>
         </div>
       {% endif %}
+      {% if not disable_form_login %}
       <div class="form-group">
         <label class="form-label" for="{{ form.username.id_for_label }}">Username</label>
         {{ form.username|add_class:'form-input'|attr:'placeholder: ' }}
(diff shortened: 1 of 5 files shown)


sissbruecker:
Reading through #730 again, I don't think the intention is to hide the login form. The goal is to automatically initiate the OIDC flow when opening the app without having to click a button on the login page.

sissbruecker:
I don't see much value in this change, so I'll close the PR for now.

Issue #205, opened 2022-02-22
Options to make tag panel more managable


gingerbeardman:
With over 4000 bookmarks I have a lot of tags, it would be useful to be able to:

- show only those used more than a certain number of times (Pinboard "top tags", etc)

or
- hide the list

sissbruecker:
See also #594

PR #975, opened 2025-02-02
Add option to collapse side panel

sissbruecker:
Related to https://github.com/sissbruecker/linkding/issues/205


marinkaberg:
Option **Collapse side panel** in **Settings** has a side effect: **Notes** element of all bookmarks stops open.

sissbruecker:
@marinkaberg I couldn't reproduce this from some quick testing. Could you elaborate a bit more? Like, which steps you took, which browser you are using an so on.


https://github.com/user-attachments/assets/09440a4d-0908-41fe-97eb-63d5fbb53180

sissbruecker:
No worries, maybe some caching issue.

Issue #952, opened 2025-01-20
Feature Request:  function to reprocess existing links to add missing description


manilx:
I imported links from linkwarden (as html). Unfortunately the Description field is empty.

I would be great if there was a way to revisit all links and add the missing description. It's being filled on newly created entries.

PR #979, opened 2025-02-02
Adds the option to fetch metadata to imports

daniel-endraws:
I had the same request as #952, so I implemented it in case anyone else does too!

Quickly testing with 177 imported bookmarks (all with missing fields) took ~20 seconds (85 seconds without the threading).

### DIFF
--- a/bookmarks/services/bookmarks.py
+++ b/bookmarks/services/bookmarks.py
@@ -69,7 +69,7 @@ def update_bookmark(bookmark: Bookmark, tag_string, current_user: User):
     return bookmark
 
 
-def enhance_with_website_metadata(bookmark: Bookmark):
+def enhance_with_website_metadata(bookmark: Bookmark, save: bool = True):
     metadata = website_loader.load_website_metadata(bookmark.url)
     if not bookmark.title:
         bookmark.title = metadata.title or ""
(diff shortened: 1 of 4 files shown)


bookmarks/services/importer.py
```diff
@@ -169,6 +172,18 @@ def _import_batch(
             logging.exception("Error importing bookmark: " + shortened_bookmark_tag_str)
             result.failed = result.failed + 1
 
+    if options.get_metadata_flag:
+        bookmarks_missing_metadata = filter(
+            lambda b: not b.title or not b.description,
+            bookmarks_to_create + bookmarks_to_update,
+        )
+
+        with ThreadPoolExecutor(5) as executor:
```
sissbruecker, line 181: This should be a background task instead, you can check `bookmarks/services/tasks.py` as a reference for existing ones. So you could add a new task there, then schedule one per bookmark from here.

Tests would be welcome as well, `test_bookmarks_tasks.py` is used for testing tasks themselves, and `test_importer.py` could test that the task is called on import.

bookmarks/services/bookmarks.py
sissbruecker, line 1: This can be reverted when adding the background task.

sissbruecker (commented):
Thanks, this is a good start. Added some comments below.

sissbruecker:
Closing as stale. If there is interest in proceeding with this, see https://github.com/sissbruecker/linkding/pull/999 for an example of an async task that updates bookmark metadata.

PR #1041, opened 2025-04-06
Search on whole tags instead of start for tag autocomplete

agateblue:
I'm finding the current tag autocomplete a bit strict when searching. Assuming I have the following tags in database:

- `animal/python`
- `language/python`

Typing `python` in the input will show no suggestion. This is because the search matches from the start of the tag name. This patch changes the behaviour to search in the whole tag name, and not only at the start.

# Current behaviour

image

# New behaviour

image


sissbruecker:
I'd prefer to make this optional. Personally I only use single words as tags and matching from the start limits the results better in that cases.

sissbruecker:
Closing as stale

PR #1050, opened 2025-04-19
Add option to import description as notes

sonicdoe:
Adds a second import option to import a bookmark’s description as notes:

This makes it easier to import bookmarks exported by other services such as Pinboard or Raindrop.io which store the user’s notes in the description field (as the Netscape file format doesn’t provide a dedicated notes field).

### DIFF
--- a/bookmarks/services/importer.py
+++ b/bookmarks/services/importer.py
@@ -23,6 +23,7 @@ class ImportResult:
 @dataclass
 class ImportOptions:
     map_private_flag: bool = False
+    description_as_notes: bool = False
 
 
 class TagCache:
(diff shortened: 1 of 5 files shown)


sissbruecker:
Can you elaborate what problem you're trying to solve? If the notes from those apps end up in the description, what is the problem with that?

One thing I noticed is that pinboard now adds a blockquote tag around every description, which was not the case when I exported my data from there. But otherwise it looks like you can not add custom tags or markdown to the description, so maybe some sanitization to remove those wrapper tags would be more useful then.

sissbruecker:
Closing as stale

PR #1060, opened 2025-05-08
Allow pre-filling tags in new bookmark form

dasrecht:
Hi there 👋 

I'm moving from another comparable system to linkding and realized that I can't set the bookmarks in custom bookmark snippets. This PR reads `tag_string` from the GET requests, allowing to pre-fill tags.

Hope that is ok to add it like I've done in the PR. Let me know if there's also some documentation that needs to be amended.

/dasrecht

### DIFF
--- a/bookmarks/forms.py
+++ b/bookmarks/forms.py
@@ -41,6 +41,7 @@ class BookmarkForm(forms.ModelForm):
                 "title": request.GET.get("title"),
                 "description": request.GET.get("description"),
                 "notes": request.GET.get("notes"),
+                "tag_string": request.GET.get("tag_string"),
                 "auto_close": "auto_close" in request.GET,
                 "unread": request.user_profile.default_mark_unread,
             }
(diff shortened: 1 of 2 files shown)


sissbruecker:
Makes sense, changed the parameter name to `tags` though.

jk:
By the way, I've modified the bookmarklet to also pre-fill a couple of the other fields:

```javascript
javascript:(function () {
    let w = 580;
    let h = 780;
    let l = Math.floor((screen.width - w) / 2);
    let t = Math.floor((screen.height - h) / 2);
    let u = encodeURIComponent(location.href);
    let s = encodeURIComponent(getSelection().toString());
    let ti = encodeURIComponent(document.title);

    let lnk = `https://linkding.example.tld/bookmarks/new?url=${u}&description=${s}&title=${ti}&auto_close#id_tag_string`;

    window.open(lnk, "Bookmarklet", `width=${w},height=${h},left=${l},top=${t}`);
})();
```

Issue #940, opened 2025-01-08
Allow `date_added` to be specified in `POST /api/bookmarks/` API


jmason:
Hi!

I'm working on syncing my Pinboard bookmarks to a Linkding server, and have run into a minor issue.  My Pinboard bookmarks have an existing timestamp and I'd like to preserve that, but the `POST /api/bookmarks/` API does not allow `date_added` or `date_modified` to be set.

I'm working around this by using the Netscape bookmarks importer and an intermediate HTML file, but that's suboptimal.

Would you be open to a PR to made those 2 fields non-read-only?  Cheers :)

jmason:
Does that PR fix this issue? I'm not seeing it.... but I could be missing something

PR #1063, opened 2025-05-13
Make date_added and date_modified optionally writable fields for the POST /api/bookmarks/ API

jmason:
As discussed at https://github.com/sissbruecker/linkding/issues/940, syncing bookmarks between platforms is a good use case where it makes sense to be able to POST a bookmark via the API with a specified date_added and/or date_modified timestamp set.  This PR makes them optional parameters on that API.

This resolves https://github.com/sissbruecker/linkding/issues/940 .

### DIFF
--- a/bookmarks/api/serializers.py
+++ b/bookmarks/api/serializers.py
@@ -53,8 +53,6 @@ class BookmarkSerializer(serializers.ModelSerializer):
             "favicon_url",
             "preview_image_url",
             "tag_names",
-            "date_added",
-            "date_modified",
             "website_title",
             "website_description",
         ]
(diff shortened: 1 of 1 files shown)


jmason:
hey folks, any chance of a review of this one?

bookmarks/api/serializers.py
```diff
@@ -111,6 +112,17 @@ def create(self, validated_data):
...
+
+        # override default timestamps, if specified
+        added = 'date_added' in validated_data
+        modified = 'date_modified' in validated_data
+        if added or modified:
+            if added:
+                saved_bookmark.date_added = validated_data['date_added']
+            if modified:
+                saved_bookmark.date_modified = validated_data['date_modified']
+            saved_bookmark.save()
```
sissbruecker, line 124: Instead of saving the bookmark twice it would probably make more sense to change `create_bookmark` so that it only initializes the dates if the new bookmark doesn't already specify them.

Adding some test cases would also be helpful.

jmason: yes, fair enough.  Will do!

sergeydubak:
@jmason hi, are you going to work on this?
If not, I can try to suggest another PR

jmason:
@sergeydubak yes please!  I have been meaning to get around to it, but haven't got around to it :(

jmason:
I've finally gotten around to updating this; it should do the trick now.... reviews would be great :)

PR #1075, opened 2025-05-25
Add Trafilatura archiver as lightweight alternative to SingleFile for HTML snapshots

domage:
### Summary
SingleFile have issues working on machines with restricted memory (1Gb) and handling complex pages with a lot of embedded media. This PR introduces Trafilatura as a lightweight alternative archiver for creating HTML snapshots, providing a pure Python solution that doesn't require browser dependencies like Chromium.

You may check this build from:
[docker pull domage/linkding:test](https://hub.docker.com/repository/docker/domage/linkding/general)

WARNING: Despite the fact that the image _should_ work without any problems with existing installations, remember that **this is a test build, so make a copy of your data before you deploy, or deploy in a separate instance.** 

### Key Features
- Lightweight archiver: Pure Python implementation using Trafilatura library
- Image embedding: Automatically embeds images as base64 data URLs with configurable size limits
- Video handling (very basic): Converts _some_ video embeds to clickable links for better accessibility (not perfect, but handles some cases)
- Backward compatibility: Maintains full compatibility with existing SingleFile installations
- Memory efficient: Significantly lower resource usage compared to browser-based solutions (works well on 1Gb Mem instances)

### New Components
- trafilatura_extractor.py: Core extraction engine with content processing
- Enhanced assets.py: Archiver selection logic with fallback mechanism
- Updated contexts.py: UI integration for snapshot button visibility in base image

New environment variables for Trafilatura configuration:
```
LD_ARCHIVER_TYPE=trafilatura|singlefile  # Choose archiver (default: singlefile)
LD_TRAFILATURA_TIMEOUT_SEC=30           # Request timeout
LD_TRAFILATURA_EMBED_IMAGES=True        # Embed images as base64
LD_TRAFILATURA_MAX_IMAGE_SIZE=1048576   # Max image size (1MB)
LD_TRAFILATURA_USER_AGENT=linkding/1.0  # Custom user agent
```

### Backward Compatibility
- No breaking changes: Existing installations continue using SingleFile by default
- Seamless migration: Switch archivers via environment variable
- Same output format: Both archivers produce compatible .html.gz files

Test run:
```
----------------------------------------------------------------------
Ran 814 tests in 176.715s

OK
Destroying test database for alias 'default' ('file:memorydb_default?mode=memory&cache=shared')...

```

### DIFF
--- a/bookmarks/services/assets.py
+++ b/bookmarks/services/assets.py
@@ -26,8 +26,37 @@ def create_snapshot_asset(bookmark: Bookmark) -> BookmarkAsset:
         content_type=BookmarkAsset.CONTENT_TYPE_HTML,
         display_name=f"HTML snapshot from {timestamp}",
         status=BookmarkAsset.STATUS_PENDING,
-    )
+    )    
     return asset
+
+
+def _try_trafilatura_snapshot(url: str, filepath: str) -> bool:
+    """Try to create snapshot using Trafilatura. Returns True on success."""
+    try:
+        import trafilatura
+        logger.info(f"Creating snapshot with Trafilatura for {url}")
+        trafilatura_extractor.create_snapshot(url, filepath)
+        return True
+    except ImportError:
+        logger.warning("Trafilatura not available")
+        return False
+    except Exception as e:
+        logger.warning(f"Trafilatura failed for {url}: {e}")
+        return False
+
+
+def _try_singlefile_snapshot(url: str, filepath: str) -> bool:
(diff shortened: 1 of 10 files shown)


sissbruecker:
Thanks for your efforts. I gave this a shot. While this looks interesting, I don't think the output is what most users would expect, plus it has the downside that it can't capture something meaningful from SPAs. For me the downside would be to have to maintain two different versions of the feature. As such I'd prefer not go ahead with this at this point.

PR #1079, opened 2025-05-28
Add CloudBreak on Managed Hosting

benjaminoakes:


sissbruecker:
Thanks, though I'd prefer to only add this once the service actually launches.

benjaminoakes:
Totally understood.

We now have an instance running and have transitioned from pre-orders to subscriptions.

We'd also like to provide the same "10% of profits" arrangement we have with other projects.  I've noted that you have a [disclosure](https://linkding.link/acknowledgements/#pikapods) for PikaPods.  If there is anything that we need to handle on our end, please let us know.  Otherwise, we'll be in touch as needed.

Thank you for making Linking!  We appreciate all the work that has gone into it.

benjaminoakes:
@sissbruecker Anything else you'd like to see?

sissbruecker:
Nope, looks good.

PR #1083, opened 2025-06-02
Add exclusion of Tags and Terms to Search

AerrowShadow:
Adding a "-" before tag or term in the search means that bookmarks containing it will not be in the search results.
Tag cloud has a button to allow to choose  whether a clicked tag should be in the search results or excluded.

### DIFF
--- a/bookmarks/frontend/components/SearchAutoComplete.svelte
+++ b/bookmarks/frontend/components/SearchAutoComplete.svelte
@@ -102,6 +102,17 @@
           tagName: tag.name
         }))
     }
+    else if (currentWord && currentWord.length > 2 && currentWord[0] === '-' && currentWord[1] === '#') {
+      const searchTag = currentWord.substring(2, currentWord.length)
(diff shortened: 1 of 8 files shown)


sissbruecker:
Thanks, I'm not sure about the UX in the tag cloud, seems quite confusing. I'll have to think about this a bit.

There's now also https://github.com/sissbruecker/linkding/pull/1097, which allows to exclude tags as part of a search preset at least, though not through the search field.

AerrowShadow:
Yeah the tag cloud UX I implement is not the best.

My first Idea of different actions for one Button(Tag), one action happens on left click the other on right click, I had no idea how to implement.
So I thought of two other ways:
-The Button to switch whether a click on a tag cloud tag counts as exclude or include.
-And the other not chosen idea was two tag clouds each in its own tab one for exclude one for include.
And now that I think about it as Text for the Button exclude and include would probably have been better.

sissbruecker:
Closing in favor of https://github.com/sissbruecker/linkding/pull/1198

PR #1094, opened 2025-06-14
Date-based hierarchical asset dir

hkclark:
I have over 50,000 bookmarks and storing that many in a single asset directory and storing that many asset files in a single directory doesn't seem ideal.  This PR adds to `_generate_asset_filename()` the feature where it add `{year}/{month}` to the front of the filename and creates the directory/directories if they don't exist.  In all of my testing it's backward compatible... it leaves any existing assets as-is and just starts putting new assets into the "hierarchical" structure.  The tests have been updated as well.

Thank you

### DIFF
--- a/bookmarks/services/assets.py
+++ b/bookmarks/services/assets.py
@@ -164,8 +164,15 @@ def _generate_asset_filename(
         else:
             return "_"
 
+    year = asset.date_created.strftime("%Y")
+    month = asset.date_created.strftime("%m")
     formatted_datetime = asset.date_created.strftime("%Y-%m-%d_%H%M%S")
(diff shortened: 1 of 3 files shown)


huyz:
Nice, that's useful

hkclark:
@sissbruecker I just wanted to check on this PR and see if there was anything I could do on my end to have this be something you would consider merging into main?  I have been running this on my instance since mid-June and it has been working well.  It shouldn't impact any existing assets... they stay as is... it just adds the subdirectories to any new entries and helps prevent having too many files in a single directory.  I have found it very useful and thought others might as well.  Thanks & regards.

bdwilson:
agreed this would be useful, but it looks like there are lots of PR's that haven't been accepted :(

PR #1120, opened 2025-07-04
Fix: Prevent layout issues caused by long bookmark title in detail view

WooHooDai:
This PR fixes layout issues on the bookmark detail (view) page caused by excessively long titles.
## Problem
1. **Overlong Titles**:
When a bookmark is created via the RESTful API, if the parsed title exceeds 512 characters, it is not limited, causing the detail view to break the layout.
(See attached image 1)
2. **URL as Title**:
When a URL is used as the title, it does not wrap, resulting in layout overflow.
(See attached image 2)
## Solution:
1. Limit the maximum length of bookmark titles displayed in the detail view.
2. Ensure that long titles and URLs will wrap appropriately to prevent layout overflow.
## Screenshots:
- Problem:
Image 1: image
Image 2: image

- Fixed:
image
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -196,3 +196,6 @@ typings/
 /chromium-profile
 # direnv
 /.direnv
+
+.DS_Store
+PROJECT.md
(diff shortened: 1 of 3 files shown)


sissbruecker:
Thanks for pointing that out. Wasn't quite happy with the implementation, so I created https://github.com/sissbruecker/linkding/pull/1150 as an alternative.

Issue #869, opened 2024-10-02
Bulk edit for "Create html snapshot"


cpresse:
After applying the options from issue #690 for creating snapshots of sites like reddit it would be great to have the option create new html snapshots by batch via the bulk edit function.

PR #1132, opened 2025-07-18
Bulk create HTML snapshots

Tql-ws1:
Closes https://github.com/sissbruecker/linkding/issues/869

### DIFF
--- a/bookmarks/templates/bookmarks/bulk_edit/bar.html
+++ b/bookmarks/templates/bookmarks/bulk_edit/bar.html
@@ -23,6 +23,7 @@
           <option value="bulk_unshare">Unshare</option>
         {% endif %}
         <option value="bulk_refresh">Refresh from website</option>
+        <option value="bulk_snapshot">Create HTML snapshot</option>
       </select>
       <div class="tag-autocomplete d-none" ld-tag-autocomplete>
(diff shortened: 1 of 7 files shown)


bookmarks/templates/bookmarks/bulk_edit/bar.html
```diff
@@ -23,6 +23,7 @@
           <option value="bulk_unshare">Unshare</option>
         {% endif %}
         <option value="bulk_refresh">Refresh from website</option>
+        <option value="bulk_snapshot">Create HTML snapshot</option>
```
sissbruecker, line 26: This is only supported in the plus image, so it would be better to only show this if the feature is actually active. Here is a check that implements this when scheduling a background task: https://github.com/sissbruecker/linkding/blob/887e42a9fbc35f8ea0b5f7e3fca77b729f280dca/bookmarks/services/tasks.py#L254-L255

To make this available in the template I would suggest to expose `settings.LD_ENABLE_SNAPSHOTS` in a field in `BookmarkListContext`.

Would also be good to have tests to verify that the option is only shown when the feature is enabled.

Tql-ws1:
Sorry, I recently switched from using git to [jujutsu](https://jj-vcs.github.io/jj/latest/), and the force push above should have been pushed to a branch in my fork repository.

PR #1138, opened 2025-07-23
Redirect supervisord output to stdout in docker

### LINKED ISSUE #1123: Send all logs to docker's log output
I'm running linkding under docker-compose. I was trying to debug why background tasks weren't working properly and it took me a while to discover the `/etc/linkding/background_tasks.log` file within the container that had the relevant error logs.

Can the container be configured to redirect the `background_tasks.log` and `supervisord.log` logging to go to the container's stdout instead? That would simplify debugging when running under docker and it would also follow common practice in the docker world of forwarding logs from the container to docker for aggregation.

garyp:
When running services in docker it's usually most convenient to use docker's built-in log management that collects logs from container stdout. Having logs go to log files inside the container is much more cumbersome to work with.

Fixes #1123

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -29,6 +29,8 @@ chown -R www-data: /etc/linkding/data
 
 # Start background task processor using supervisord, unless explicitly disabled
 if [ "$LD_DISABLE_BACKGROUND_TASKS" != "True" ]; then
+  # when running under Docker send the logs to stdout so that Docker can collect them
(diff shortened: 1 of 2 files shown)


sissbruecker:
Did some more testing and this can not be solved by just redirecting the log output. supervisor needs to run as main process that manages both the web server and the background task processor, which then allows to capture output from both to the container logs. 

I've added an experimental environment variable to allow starting supervisor as the main process in https://github.com/sissbruecker/linkding/pull/1270. I'm not 100% certain about the implications and a proper setup for this, so for now this can be used to test this setup.

PR #1146, opened 2025-08-05
Use filename when downloading asset through UI

proog:
Currently, the asset view response carries no information about the name of the asset. This is fine for viewing HTML snapshots in the browser, but downloading the asset defaults to a file name consisting of just the asset id (with no extension). This is especially cumbersome if the browser doesn't support displaying the file type itself - the browser might attempt to download the file directly without prompting for a file name, and the user will have to manually rename the file afterwards.

The asset download route in the API already handles this by using the display name of the asset in the Content-Disposition header. This PR reuses the same logic in the assets view, but with an "inline" disposition type so the browser will still attempt to display the file instead of downloading it. If the browser/user attempts to save the file to disk, the downloaded file now uses the display name by default.

### DIFF
--- a/bookmarks/api/routes.py
+++ b/bookmarks/api/routes.py
@@ -199,13 +199,10 @@ class BookmarkAssetViewSet(
                 if asset.gzip
                 else open(file_path, "rb")
             )
-            file_name = (
-                f"{asset.display_name}.html"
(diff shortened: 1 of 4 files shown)


sissbruecker:
Looks good, thanks!

PR #1148, opened 2025-08-08
Handle bare domain URLs

### LINKED ISSUE #1141: Feature Request: Automatically handle bare domain names
Currently if I try to add a bookmark using something like "test.com", it returns an error as an invalid URL. It would be very helpful for setting up initial bookmarks to handle URLs that are missing the protocol, either by adding it automatically or accepting the domain and normalizing it.

If you like this feature and will accept it I would be happy to pick it up! I am taking an Open Source Software class and will need to implement an actionable issue for a later assignment.

Coneco629:
Fixes #1141 

This change allows users to enter bare domain names (e.g example.com) when adding or editing bookmarks. If no scheme is provided, "https://" is prepended before validation and saving.

- Updated BookmarkURLValidator to normalize bare domains before validation
- Updated BookmarkForm.clean_url to save the normalized value
- Restricted to http and https schemes only

### DIFF
--- a/bookmarks/forms.py
+++ b/bookmarks/forms.py
@@ -78,12 +78,28 @@ class BookmarkForm(forms.ModelForm):
             return create_bookmark(bookmark, tag_string, self.request.user)
 
     def clean_url(self):
+        from urllib.parse import urlparse
+        raw = (self.cleaned_data.get("url") or "").strip()
+
+        # If user types a bare domain like "example.com"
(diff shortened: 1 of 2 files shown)


sissbruecker:
Taking a closer look there are some edge cases missing from the logic, and it would also require adding support for other paths in the application such as loading website metadata based on the URL field input. Let me close this for now.

PR #1161, opened 2025-08-16
Document API bundle filter

proog:
It's possible to filter bookmarks in the API by providing a bundle id, but this is not mentioned in the docs.


sissbruecker:
Thanks!

Issue #42, opened 2020-12-28
Add an invitiation workflow


sissbruecker:
Currently there is no UI workflow to add new users to the system. The only options are creating new super-users using the Django script or enabling registration for a limited time. For self-hosted scenarios, it would make sense if the application would support invites. That would make it simple to add specific users (such as family members, friends, colleagues) without setting a password for them.

The workflow would look something like this:
- Application provides a page + form to invite a new user per mail
- User clicks on link in mail and lands on page / form where they can complete their registration details and set their password
- Completing the form enables the account and the user is now able to login

Would need to consider if this needs to be behind a config flag or if the invite feature can be active all the time (for super users).

sissbruecker:
Rather than sending a mail automatically, it would be a lot easier to just provide an invite link that can be shared however the user likes. If you just want to invite like 3 people, sharing the 3 links is likely faster than having to figure out how to configure an SMTP service.

Lanhild:
Agreed, as an admin, having the ability to create an invitation link that can expire in `x` hours, days... or after a specific number of uses. Very useful in large-scale use cases

PR #1162, opened 2025-08-16
Fix error when filtering bookmark assets in the admin UI

proog:
Executing any search on `/admin/bookmarks/bookmarkasset/` results in a HTTP 500:

```
django.core.exceptions.FieldError: Cannot resolve keyword 'custom_display_name' into field. Choices are: asset_type, bookmark, bookmark_id, content_type, date_created, display_name, file, file_size, gzip, id, latest_snapshot, status
```

It seems that calculated display names are not supported as search fields. This PR changes the search field back to `display_name` to make searching work again. Unfortunately, this means that searching for a synthetic display name like `Bookmark Asset #42` is not supported.

### DIFF
--- a/bookmarks/admin.py
+++ b/bookmarks/admin.py
@@ -214,7 +214,7 @@ class AdminBookmarkAsset(admin.ModelAdmin):
 
     list_display = ("custom_display_name", "date_created", "status")
     search_fields = (
-        "custom_display_name",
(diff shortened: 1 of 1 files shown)


sissbruecker:
Thanks! Makes sense that this is not searchable, also I don't see how there would not be a display name.

PR #1186, opened 2025-09-17
Fix pagination links to use relative URLs

dunlor:
When I run the application in docker container and access it behind reverse proxy, I see all pagination links point to "localhost:8000" . 
This change fixes the absolute paths to relative (without host and port).


sissbruecker:
Looks reasonable, thanks!

PR #1187, opened 2025-09-17
Fix queued tasks link when context path is used

dunlor:
Currently the link is hardcorded to /admin/tasks/, which is incorrect in case LD_CONTEXT_PATH  is not empty.


sissbruecker:
Thanks!

PR #1195, opened 2025-10-03
Added Javascript client and library for Linkding REST API

vbsampath:
Added Javascript client (linkding-client) and library (linkding-api) for Linkding REST API


sissbruecker:
Thanks, I removed the client repo as it's just usage examples that are also reachable from the api repo.

PR #1208, opened 2025-10-15
Fix devcontainer

m3eno:
- Update Python version to 3.13
- Install uv
- Update `postCreateCommand` to use uv
- Update README with uv commands


sissbruecker:
Looks good, thanks!

PR #1216, opened 2025-10-27
Bump supervisor to 4.3.0 to fix warning

simonhammes:
This was the warning:

```
/etc/linkding/.venv/lib/python3.13/site-packages/supervisor/options.py:13: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
```


sissbruecker:
Thanks!

PR #1241, opened 2025-12-11
Disable bulk execute button when no bookmarks selected

emanuelebeffa:
## Summary
- Disable execute button in bulk edit mode when no bookmarks are selected

## Changes
- Added `updateExecuteButton()` method to toggle button state based on checkbox selection
- Execute button is now disabled by default and only enabled when at least one bookmark is checked
- Button state updates automatically when toggling individual checkboxes or when using the select all action


sissbruecker:
Thanks!

Issue #492, opened 2023-06-19
Feature idea: webhooks


pettijohn:
I was looking into self-hosted options that take web snapshots similar to Internet Archive Wayback Machine ([see a few options here](https://github.com/awesome-selfhosted/awesome-selfhosted#archiving-and-digital-preservation-dp)), and thinking how that could be integrated with Linkding. A generic solution could be to add the ability to add webhooks that Linkding triggers upon saving a new bookmark. Configuration would allow end user to pick HTTP method (GET/POST at least) and a URL with tokens of at least bookmarked url, comma separated list of tag; title & description maybe, and possibly a user identifier for who initiated (but really that should be handled by the remote service with an API key or similar). POST methods might need another textarea input to construct the body.

Not sure if there is any demand for such a feature, but I thought I'd share the idea. Personally, I like the idea of being able to trigger services that export a web page to PDF or MHTML, or perhaps trigger youtube-dl for links with certain tags, storing those in my own personal archive.

sissbruecker:
I'm not sure that a generic mechanism provides a good user experience here. Then everyone who would want to configure some service would have to start from scratch in figuring out which URL, headers etc. to configure. Looking at Miniflux for example they support a specific set of services, each with their own configuration UI, and building the request just happens under the hood.

Another aspect is surfacing links in the UI. If you configure a custom archiving service, you would probably expect that the bookmark list shows links to the archived snapshots that have been created through that. That would just be another hassle if you would have to configure those manually, even if it's just a name for the service.

pettijohn:
I take your point, especially showing the link to archived snapshot inline. Does linkding have a plugin/extension architecture to allow third parties to add functionality without modifying the linkding source code?

mzehrer:
I second the the wish for a general purpose webhook to publish lifecycle events (CRUD) for bookmarks. Linkding is a central part in my knowledge retrieval  workflows and I would love to be able to integrate it with my other applications. I would be happy with just sending out the bookmark data to a http endpoint.

voltagex:
Just found this project and issue. I would also definitely use a webhooks feature to be able to take actions when certain links or domains are saved

voltagex:
https://github.com/danihodovic/django-webhook/ might be an option

gusvd:
Is there a workaround to monitor changes in linkding? Webhooks would be ideal. I know I could use the API with the date filters, but in case of deletion, I'll need to retrieve and compare the whole base.

[...]

PR #1244, opened 2025-12-12
Feat: webhook support for added, edited and deleted bookmarks

pava0:
Hi everyone! New contributor here!

This PR addresses the feature idea #492 for a simple POST webhook to send the newly added/updated/deleted bookmark to other self-hosted services like crawlers (The Wayback Machine, Browsertrix), notification servers (Gotify, ntfy.sh) or worfklow automation platforms (n8n, Node-Red).

At the moment this feature allow the user to receive in the body of the request the event type ("new_bookmark", "updated_bookmark" or "deleted_bookmark") and the relevant data (url, id, owner username, title, description, notes, website title, website description, date added and tags).
This happens only for the bookmarks that have a user specified tag (e.g. "crawl").

I've also added a simple username:password authentication method for webhooks served behind credentials.

image

At the moment it's just a simple POST request.
If @sissbruecker thinks it is worth it to expand this feature further, these are possible improvements:

- Choice between POST requests (with bookmark's data in the body) and GET request with data encoded in the URL
- User configurable body content for POST request (select which fields to include and how to name them)
- List based tag-webhook system, to allow different tags to send requests to different webhooks (one for youtube-dl, one for page s to archive, etc)
- Add a button on the view page for the bookmark to manually send the URL to a specific endpoint.

Let me know what you think, and thank you for Linkding! I've been using it regularly for more than a year


gusvd:
This would be super beneficial for my use-case. Future improvement would be to also send webhooks for deleted and modified entries.

pava0:
> This would be super beneficial for my use-case. Future improvement would be to also send webhooks for deleted and modified entries.

Updated the PR to push the webhook for added, updated or deleted bookmarks.

nkoehring:
One thing that would be super useful is to filter for shared/public bookmarks, so users could make sure only their public bookmarks are sent to other services or (like in my case) their website generator.

Issue #1072, opened 2025-05-22
Support for PDF Archives?


johnmaguire:
I love Linkding - great for bookmarks and keeping a copy of articles / guides in case they go missing later.

Sometimes what I'm bookmarking isn't an HTML page, but a PDF. In these cases, the archive function does not work correctly. I manually upload the PDF (thanks for the manual upload functionality!) but it'd be great if I didn't have to.

enncoded:
What is the manual upload you're referring to?

johnmaguire:
@enncoded Click View on a link then under Files you should see an Upload file option.

AdamVig:
I forked https://github.com/proog/linkding-media-archiver to https://github.com/AdamVig/linkding-pdf-archiver for this purpose. I've been using it for a bit now and it seems to do the trick!

PR #1271, opened 2026-01-06
Download PDF instead of creating HTML snapshot if URL points at PDF

sissbruecker:
Closes https://github.com/sissbruecker/linkding/issues/1072


johnmaguire:
Thank you!! :)

pettijohn:
Linkding keeps getting better! THANK YOU!

PR #1273, opened 2026-01-08
Fix cookie path for non-default context path

zhangyoufu:
Cookie path did not respect `LD_CONTEXT_PATH`.
Sensitive cookies like `csrftoken` and `sessionid` may be leaked to other applications hosted on the same origin.


zhangyoufu:
Hi @sissbruecker, when you have a moment, could you please take a look at this PR? Thanks!

ghost:
Hi @zhangyoufu 

What’s your use case for running Linkding on a subpath? I’m asking because it might be worth considering whether we can remove `LD_CONTEXT_PATH` altogether, since it adds some additional complexity and overhead.

zhangyoufu:
> What’s your use case for running Linkding on a subpath?

The use case is exactly “running Linkding on a subpath” so that I don’t have to setup another subdomain for it. It’s viable and supported so I did it.

> it might be worth considering whether we can remove LD_CONTEXT_PATH altogether

+1

> since it adds some additional complexity and overhead.

Blame Django for this messy implementation. My point is that web browsers does not provide isolation between subpath. It’s generally safer to deploy linkding in its own origin, not share the same origin with other apps.

sissbruecker:
Thanks, I was not able to push to your branch so I added your changes + some cleanup in 0c2362b62a3832a0fe6810ec8cabdc3e209aea85.

PR #1295, opened 2026-02-01
Allow configuring session duration

### LINKED ISSUE #1226: [Feature] Setting for duration of session
Hi
can you add a setting for the duration of a session? At the moment the cookie will expire after 2 weeks

fardog:
Allows overriding [`SESSION_COOKIE_AGE`](https://docs.djangoproject.com/en/6.0/ref/settings/#session-cookie-age) via an environment variable, to support setting a longer session expiry when hosting in docker. The provided default value is the current Django default.

resolves #1226


subdavis:
I just made this change also, but found your Pr before opening mine. Needs to be added to .env.sample

would really appreciate this feature, thanks.

PR #1297, opened 2026-02-03
Add note for using auth-proxies with rest-clients

### LINKED ISSUE #1296: CRSF error when creating bundle if submitting after preview updates when using Caddy and forward auth
The following video demonstrates the issue, caused by my Caddy config but possibly linkding could have a more graceful way to handle 403 in-page API calls, or the documentation could use a note.

First I create a bundle, and filter by tags, **allow the preview to update**, (note the 403 response in the network log) then create the bundle, which results in a CSRF error.

Then (after refreshing) I create another bundle, filter by tags but **submit the form** **before** **the preview can update** and the bundle is created.

https://github.com/user-attachments/assets/d074db42-201a-4c3d-972e-775a20829267

I believe this behaviour is caused by in-page requests to the API failing (silently), and invalidating the CSRF token (?).

---

I had the following in my caddy configuration, where calls to `/api/` are not passed to forward auth (as they are coming from non-browsers that were not logged in via authelia):

```Caddyfile
site { 
 handle /api/* {
    reverse_proxy linkding:9090
  }

  handle {
    import authelia-forward-auth
    reverse_proxy linkding:9090
  }
}
```

In my case I can fix this by a more explicit matcher for caddy, only applying the forward auth bypass if an `Authorization Bearer*` header is present.

```caddyfile
site {
  @external_api_call {
    path /api/*
    header Authorization Bearer*
  }

  handle @external_api_call {
    reverse_proxy linkding:9090
  }

  handle {
    import authelia-forward-auth
    reverse_proxy linkding:9090
  }

}
```

Probably this is the same root as https://github.com/sissbruecker/linkding/issues/523

rktjmp:
Closes #1296


rktjmp:
I put this in "how to" but now I think it should maybe be in "installation -> reverse proxy setup"? I missed that heading before.

sissbruecker:
Thanks, the in-page carve out rule is helpful. I've moved this as a note to the option that enables auth proxies and also added a troubleshooting section that explains about the issue.

PR #1300, opened 2026-02-10
Fixing weird underlined whitespace in tag cloud

salguerooo:
I noticed today that sometimes a tag in the tag cloud has some whitespace which is included in the a tag which also gets underlined when hovering it.

This is what it used to look like:
Screen Shot 2026-02-09 at 17 24 55

This is what it looks like after the fix:
Screen Shot 2026-02-09 at 17 25 53

The way I fixed it is by using [spaceless](https://docs.djangoproject.com/en/6.0/ref/templates/builtins/#spaceless) to remove any whitespace in between html tags. I ended up just wrapping the whole template since it was affecting the selected tags as well.


sissbruecker:
It looks like this solution prevents tags from wrapping when they overflow, which currently relies on whitespace. Created https://github.com/sissbruecker/linkding/pull/1437 as an alternative.

PR #1306, opened 2026-02-20
docs: Add Hostim.dev to the managed hosting providers list.

pavel1337:
Adds Hostim.dev to the Fully Managed section of the managed hosting page.

Hostim.dev is a hosting platform based in Germany (EU). Linkding is
available as a one-click template:

- Template page: https://hostim.dev/docs/templates/linkding/
- Direct deploy: https://console.hostim.dev/dashboard?preview=1&modal=1&template=linkding

The PR adds a single line under the existing PikaPods and CloudBreak
entries, matching the same format.


pavel1337:
Hi @sissbruecker, friendly bump on this one. Happy to adjust wording, move it to Self-Managed, or add a disclosure line if that fits better. Let me know what you would prefer.

sissbruecker:
Thanks for supporting linkding through your platform. I've added it as part of b7d3f3008889a6316b2702c14e8fbe51bd5d4dc6 and merged the available options into a single table.

I still think it makes sense to distinguish between a provider that provides the app to you as a service vs. a provider that allows hosting arbitrary apps (which I would quality hostim as). While your setup looks very easy to get started (and the trial without login is great) it gets a bit more technical to apply updates or set up backups. Overall I think the current docs are a fair representation otherwise and I may revisit this if I get better ideas on how to present this so that users get an idea of which level of technical expertise is required of them.

PR #1349, opened 2026-04-06
Implement search for user in public shared feed

stoeps13:
I like the shared bookmarks of a user (seen in Shaarli) very much, so I built the search into linkding.

Calling the shared feed URL without filter returns all shared bookmarks of the instance, but the call now also supports `?user=name` to filter shared bookmarks by user.

I'm not sure if this should also be mentioned in the integrations page, but for now, I'm fine knowing it's possible.

Regards,
Christoph


sissbruecker:
There were a number of smaller issues with this approach, created https://github.com/sissbruecker/linkding/pull/1436 as an alternative.

PR #1356, opened 2026-04-15
Add random sort option to bookmark search

### LINKED ISSUE #1061: Feature request - Sort by random
I have a bunch of long read articles that I like to just shuffle through when I have free time, so having a random option in the sort dropdown would be nice.

johnmaguire:
Closes #1061


patricksthannon:
Would love to see this feature added. I suffer from storing and forgetting about too many bookmarks. Reviewing by random every so often would be helpful. Luckily the api is easy to build a review script but an in app feature would be better.

PR #1357, opened 2026-04-15
Feature/auto pdf asset

RobinAngele:
Goal

This PR adds a new feature to Linkding that automatically creates a pending PDF snapshot asset for any bookmarked URL that points to a PDF file. This allows the PDF to be downloaded and stored as an asset alongside the bookmark, similar to HTML snapshots. The feature is controlled by the new LD_ENABLE_PDF_SNAPSHOTS option (default: True).
The new option is documented in docs/src/content/docs/options.md after LD_DISABLE_BACKGROUND_TASKS.

Environment

This patch was developed and tested with:

- Linkding running in Docker on a remote Debian server, using the sissbruecker/linkding:latest-plus image
- The Linkding browser extension in Brave on a development laptop
- The SingleFile extension in Brave, configured to upload snapshots to Linkding

Documentation Reference

The official documentation for server-based archiving is available at:
https://linkding.link/archiving/

It describes a setup where:

- The latest-plus Docker image enables server-side archiving
- Snapshots are generated using singlefile-cli, which runs a headless Chromium instance on the server to produce HTML files
- uBlock Origin Lite is used to block ads and unwanted content

Clarification

According to the documentation, when bookmarking a direct PDF URL, Linkding should automatically download and store the PDF as an asset.

However, this behavior did not work in our setup.

This pull request adds the missing server-side logic required for automatic PDF asset creation in this case.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -14,6 +14,8 @@ LD_SUPERUSER_NAME=
 LD_SUPERUSER_PASSWORD=
 # Option to disable background tasks
 LD_DISABLE_BACKGROUND_TASKS=False
+# Enable automatic PDF snapshot asset creation for PDF URLs (True/False)
+LD_ENABLE_PDF_SNAPSHOTS=True
 # Option to disable URL validation for bookmarks completely
 LD_DISABLE_URL_VALIDATION=False
 # Enables support for authentication proxies such as Authelia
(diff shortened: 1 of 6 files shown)


sissbruecker:
The feature was added in https://github.com/sissbruecker/linkding/pull/1271. If something doesn't work, then please open an issue with a specific sample to reproduce. But adding a second implementation of the same feature doesn't make sense.

PR #1361, opened 2026-04-21
Docs - note for nginx users on bookmark importing

ch604:
without altering the client_body_buffer_size, uploads would be restricted to around 40-50 bookmarks per file, since the default body is only 16k on 64-bit machines.

### DIFF
--- a/docs/src/content/docs/installation.md
+++ b/docs/src/content/docs/installation.md
@@ -142,6 +142,7 @@ location /linkding {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


sissbruecker:
Hard to say what your actual issue was, but the [buffer size](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_buffer_size) does not control the maximum size of the request body. There's a separate parameter for that.

PR #1375, opened 2026-05-29
Fix: Stored XSS via Reader View and SSRF via bookmark URL fetching

Yunkaiwjs:
## Summary

This PR fixes two security vulnerabilities:

### 1. Stored XSS via HTML Snapshot in Reader View (CVSS 9.6)
The  endpoint in  was rendering HTML content with the  template filter without a Content-Security-Policy header. This bypassed the CVE-2025-14202 sandbox fix that was applied to the  endpoint.

**Fix:** Added  header to the  endpoint response, matching the existing protection on the  endpoint.

### 2. SSRF via Unvalidated Bookmark URLs (CVSS 8.2)
The  and  services made HTTP requests to user-supplied bookmark URLs without validating the destination hostname. This allowed attackers to force the server to make requests to internal/private IP addresses.

**Fix:** Added  validation function that resolves the URL hostname and blocks requests to private/internal IP ranges:
- 127.0.0.0/8 (loopback)
- 10.0.0.0/8 (private)
- 172.16.0.0/12 (private)
- 192.168.0.0/16 (private)
- 169.254.0.0/16 (link-local)
- 0.0.0.0/8 ("this" network)
- ::1/128 (IPv6 loopback)
- fc00::/7 (IPv6 unique local)

The validation is applied to all three request sites:
-  - raises ValueError for blocked URLs
-  - returns None for blocked URLs
-  - returns None for blocked preview image URLs

### Files Changed
-  - Added CSP header to read() endpoint
-  - Added SSRF protection with is_safe_url()
-  - Added SSRF check for preview image URLs

### DIFF
--- a/bookmarks/services/preview_image_loader.py
+++ b/bookmarks/services/preview_image_loader.py
@@ -34,6 +34,10 @@ def load_preview_image(url: str) -> str | None:
 
     image_url = metadata.preview_image
 
+    if not website_loader.is_safe_url(image_url):
+        logger.debug(f"Blocked request to internal IP for preview image: {image_url}")
+        return None
+
     logger.debug(f"Loading preview image: {image_url}")
     with requests.get(image_url, stream=True) as response:
         if response.status_code < 200 or response.status_code >= 300:
--- a/bookmarks/services/website_loader.py
+++ b/bookmarks/services/website_loader.py
@@ -1,7 +1,9 @@
+import ipaddress
 import logging
+import socket
 from dataclasses import dataclass
 from functools import lru_cache
-from urllib.parse import urljoin
+from urllib.parse import urljoin, urlparse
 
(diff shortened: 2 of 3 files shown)


sissbruecker:
I've extracted the fix for the first issue into https://github.com/sissbruecker/linkding/pull/1402.

As for the second one, I disagree on the severity. First, it requires authentication. Furthermore linkding installations are mainly used by a single person or a trusted community, which reduces the impact a lot. The suggested fix is not exhaustive, could break legitimate usage scenarios, and can be trivially worked around with enough determination and as such gives a false sense of security. If someone really needs to isolate linkding from other services, then I feel it's best to do this at a network level.

For the future, first disclose a vulnerability privately and wait for feedback before disclosing it publicly.

PR #1378, opened 2026-05-31
Escape tag names in HTML export

### LINKED ISSUE #1377: Special characters are not escaped on export
I've just imported my bookmarks from Pinboard into Linkding, and tried to export them again to verify everything was working. But it looks like special characters in the tags are not escaped, so I got a file with a tag attribute like this:

```
TAGS="",$",$',',Bash,`,bashbook,programming,quotes,quoting,read"
```

Expected:

```
TAGS="&quot;,$&quot;,$',',Bash,`,bashbook,programming,quotes,quoting,read"
```

Working on a fix…

l0b0:
Fixes #1377.


sissbruecker:
Thanks!

PR #1411, opened 2026-07-06
Support deleting tags through the API

sethfitz:
The tags API supports list, retrieve, and create but not delete, so a tag can currently only be removed through the web UI. This adds `DELETE /api/tags/<id>/` by giving `TagViewSet` the `DestroyModelMixin` that `BookmarkViewSet` already has.

Deletion is scoped to the requesting user through the existing `get_queryset`, so a user can only delete their own tags — a tag owned by another user returns 404. Deleting a tag unassigns it from any bookmarks and leaves the bookmarks intact, matching the web UI's tag delete.

Includes tests for delete, delete of a tag that's applied to a bookmark, and the cross-user 404, plus an entry in the API reference.

---

_Drafted with Claude Code, reviewed by me before opening._

### DIFF
--- a/bookmarks/api/routes.py
+++ b/bookmarks/api/routes.py
@@ -251,6 +251,7 @@ class TagViewSet(
     mixins.ListModelMixin,
     mixins.RetrieveModelMixin,
     mixins.CreateModelMixin,
+    mixins.DestroyModelMixin,
 ):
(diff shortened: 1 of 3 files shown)


sissbruecker:
Thanks!

Issue #1327, opened 2026-03-23
Retain query string when redirecting to Shared Bookmarks "Landing Page"


karlhorky:
Similar to #1336

What are your thoughts on retaining the query string when redirecting logged-out users to the `Shared Bookmarks` option for **Landing page**:

```txt
before redirect:
https://demo.linkding.link/?q=%23china

after redirect:
https://demo.linkding.link/bookmarks/shared?q=%23china
```

That would make it easier to navigate directly to a tag in the shared bookmarks page.

Here's the view of the **Landing page** option in **Global settings**:

> **Landing page**
>
> [ Shared Bookmarks ▼ ]
>
> The page that unauthenticated users are redirected to when accessing the root URL.

PR #1416, opened 2026-07-11
Retain query string when redirecting from root view

### LINKED ISSUE #1336: Retain query string when redirecting from `/` to `/bookmarks`
Similar to #1327

What are your thoughts on retaining the query string when redirecting \*logged-in users\* from `/` to `/bookmarks`:

```txt
before redirect:
https://demo.linkding.link/?q=%23china

after redirect:
https://demo.linkding.link/bookmarks?q=%23china
```

That would make the links to a tag shorter (as well as seeming subjectively more logical)

TowyTowy:
Fixes #1336 and #1327.

The root view redirected `/` to the bookmarks page (authenticated) or the shared bookmarks landing page (unauthenticated, when configured) using a bare redirect that dropped the query string. As a result, deep links such as `/?q=%23china` silently lost their search/tag filter and landed on the unfiltered list.

This forwards the query string using the existing `redirect_with_query` helper — the same one already used by the bookmark and tag views — so `q` (and any other params) survive the redirect for both the authenticated `bookmarks.index` and the shared-landing `bookmarks.shared` cases. The login-page redirect is intentionally left unchanged, since it uses `next` rather than `q`.

Added regression tests in `test_root_view.py` covering both retaining cases plus the login redirect (which should not carry the query string). Full run: 77 passed; `ruff check` clean.

---
Disclosure: this change was prepared with the assistance of an AI agent (Claude). The code, tests, and reasoning were reviewed before submission.

### DIFF
--- a/bookmarks/tests/test_root_view.py
+++ b/bookmarks/tests/test_root_view.py
@@ -8,9 +8,17 @@ from bookmarks.tests.helpers import BookmarkFactoryMixin
 
 
 class RootViewTestCase(TestCase, BookmarkFactoryMixin):
+    def assertRedirectsToLogin(self, response, next_url):
+        # The root view redirects to the bookmarks page, which in turn
+        # redirects unauthenticated users to the login page
+        self.assertEqual(response.redirect_chain[0], (next_url, 302))
+        self.assertEqual(response.status_code, 200)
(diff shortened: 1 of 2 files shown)


sissbruecker:
Updated the implementation to also preserve the query when redirecting to the login view.

PR #1425, opened 2026-07-27
Document avoiding OIDC authentication for the API

pgaxatte:
This PR adds a section in the documentation regarding OIDC and API access: it adds a small reminder to make sure an auth proxy or a middleware does not prevent using the default token auth of the API and provides an example using traefik and docker compopse.

### DIFF
--- a/docs/src/content/docs/options.md
+++ b/docs/src/content/docs/options.md
@@ -191,6 +191,52 @@ identity_providers:
 
 </details>
 
+#### `OIDC` and API access
+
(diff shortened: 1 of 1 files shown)


sissbruecker:
Closing in favor of https://github.com/sissbruecker/linkding/pull/1297 as that had a solution that did not break API calls made by linkding itself. Also note that this issue should not apply to OIDC, which is a different auth mechanism from authentication proxies.

PR #1426, opened 2026-07-29
Fix uBlock Chromium release selection

### LINKED ISSUE #1368: [Bug] Dockerfile: uBlockOrigin download fails when latest release lacks Chromium assets
## Problem

The Dockerfile currently attempts to download uBlockOrigin using the `latest` tag. However, not every release contains a compressed package compatible with Chromium, which causes the image build process to fail.

For example, the current latest release [2026.504.2301](https://github.com/uBlockOrigin/uBOL-home/releases/tag/2026.504.2301) only includes uBOLite_2026.504.2301.safari.zip, missing the required Chromium assets.

## Suggested Solution

Instead of blindly pulling the latest tag, we can fetch the last 20 releases and filter for the most recent version that specifically contains a chromium.zip asset:

```bash
TAG=$(curl -sL https://api.github.com/repos/uBlockOrigin/uBOL-home/releases\?per_page\=20 | \
    jq -r '.[] | select(.assets[].name | contains("chromium.zip")) | .tag_name' | head -n 1) && \
    ...
```

Nameless-Monster-Nerd:
## Summary

- select the newest uBlock Origin Lite release that actually contains a Chromium ZIP
- use the asset's API-provided download URL instead of constructing one from the latest tag
- fail clearly when the API returns no compatible asset
- apply the same logic to both Docker variants and the local setup script

Closes #1368

## Validation

- `docker build --target ublock-build -f docker/default.Dockerfile .`
- live GitHub release selector returned a valid Chromium asset
- `sh -n scripts/setup-ublock.sh`
- `git diff --check`

### DIFF
--- a/docker/alpine.Dockerfile
+++ b/docker/alpine.Dockerfile
@@ -88,13 +88,14 @@ CMD ["./bootstrap.sh"]
 FROM node:22-alpine AS ublock-build
 WORKDIR /etc/linkding
 # Install necessary tools
-# Download and unzip the latest uBlock Origin Lite release
+# Download and unzip the latest uBlock Origin Lite release with a Chromium asset
 # Patch manifest to enable annoyances by default
 RUN apk add --no-cache curl jq unzip && \
-    TAG=$(curl -sL https://api.github.com/repos/uBlockOrigin/uBOL-home/releases/latest | jq -r '.tag_name') && \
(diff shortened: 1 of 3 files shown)


sissbruecker:
Adopted this into 4c755cf60a4633c3b966f70ff8a581a0ba9dacd1

PR #1438, opened 2026-08-08
test: cover missing content type from HEAD response

Iamshaibaz:


Iamshaibaz:
Adds test coverage for `detect_content_type()` when a successful HEAD response has no `Content-Type` header. This verifies the existing fallback behavior returns an empty string without making an unnecessary GET request. All 20 tests pass.

PR #1447, opened 2026-08-14
Render tag hash character with CSS so it can be hidden

### LINKED ISSUE #1408: Option to disable the Hashtag Prefix # in front of tags
Having the hashtag "hardcoded" in front of every tag is a bit annoying in the bookmark list.

It adds visual noise and coming from Pinboard I don't "need" the `#` in front of every tag.

### Suggestion
Change the implementation from to do the hashtag via CSS `::before` pseudo parameter. This way one can do either of these:

1. Add a feature toggle for the hashtag sign in the settings (on/off)
2. The hashtag sign can be handled aka disabled via **Custom CSS** rules.

sissbruecker:
Closes https://github.com/sissbruecker/linkding/issues/1408

### DIFF
--- a/bookmarks/styles/bookmark-details.css
+++ b/bookmarks/styles/bookmark-details.css
@@ -82,6 +82,10 @@
     color: var(--alternative-color);
   }
 
+  & .tags a::before {
+    content: "#";
+  }
+
   & .status form {
     display: flex;
     gap: var(--unit-2);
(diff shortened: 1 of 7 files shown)


bunchofsage:
Highly appreciate your fast implementation! But it's not yet part of a release yet, correct? It's not part of 1.46.1 according to [Releases](https://github.com/sissbruecker/linkding/releases) page

PR #1448, opened 2026-08-14
fix: replace bare except with except Exception

TrueFurina:
### Summary

Two bare `except:` clauses in settings catch everything, including
`KeyboardInterrupt`/`SystemExit`:

- `bookmarks/settings/prod.py` — fall back to a random secret key when the
  key file cannot be read
- `bookmarks/settings/__init__.py` — fall back to production settings when
  dev settings are unavailable

Replace with `except Exception:` — same intended fallback behavior, but
unexpected exceptions now propagate.

### Verification

- `python -m py_compile` on both files → OK

### Files changed

- `bookmarks/settings/prod.py` (1 line)
- `bookmarks/settings/__init__.py` (1 line)

### DIFF
--- a/bookmarks/settings/__init__.py
+++ b/bookmarks/settings/__init__.py
@@ -2,5 +2,5 @@
 # ruff: noqa
 try:
     from .dev import *
-except:
+except Exception:
     from .prod import *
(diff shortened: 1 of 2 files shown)


sissbruecker:
While this may technically be correct (haven't checked), this is not a substantial contribution and covers a case that isn't really a sensible usage scenario. Closing.

PR #1453, opened 2026-08-19
Add end-to-end tests for the bookmark & settings flows (native LinkdingE2ETestCase)

qosha1:
## What this adds

Four functional end-to-end tests (Playwright) for the bookmark and settings flows:

- **Create & view a bookmark** — fill the new-bookmark form, save, and assert the bookmark renders in the list.
- **Display settings** — change the description display option and run an import from general settings.
- **Integrations API token** — create an API token and assert the generated code is shown.
- **Change-password** — reach the change-password screen from settings and assert it renders.

They subclass `LinkdingE2ETestCase` and live under `bookmarks/tests_e2e/`, so the existing harness picks them up automatically — no config, fixtures, or dependencies added. All four pass green on a clean checkout (0 skipped, 0 flaky).

<sub>Generated with the help of <a href="https://debugg.ai">debugg.ai</a>, then reviewed and run locally before opening. Happy to adjust or drop these if you'd rather not carry them — just say the word.</sub>

### DIFF
--- a/bookmarks/tests_e2e/e2e_test_bookmarks_linkding.py
+++ b/bookmarks/tests_e2e/e2e_test_bookmarks_linkding.py
@@ -1,0 +1,53 @@
+import random
+import re
+import time
+
+from django.urls import reverse
+from playwright.sync_api import expect
+
+from bookmarks.tests_e2e.helpers import LinkdingE2ETestCase
+
+class BookmarksLinkdingE2ETestCase(LinkdingE2ETestCase):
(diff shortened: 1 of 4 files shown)


sissbruecker:
I don't think this adds much. All features are reasonably covered with Django unit tests, JS interactions are covered in the browser. Limiting E2E tests is sensible to reduce time to run tests and potential flakes. Also not happy with the borderline advertisement. Closing.

